PR #278, opened 2015-11-12
Adds a GetTF method to transform_util::Transform.

evenator:
tf::Transforms can do a lot of things that transform_util::Transforms can't, so sometimes it's convenient to just get a tf::Transform.


evenator:
Merged in, per offline discussion with @malban 
I'll issue also add this to indigo and jade.

PR #285, opened 2016-01-04
Making swri::ServiceServer's = return a value.

pjreed:
swri::ServiceServer's assignment operator was not returning a value.  This
is technically legal, and GCC will do the right thing for you, which is
to return a reference to the current object.  Clang, however, will
compile the code but generate a SIGILL exception at runtime.  This is
easily fixed by manually returning *this.


pjreed:
Note that I accidentally wrote the wrong class name in the branch name / title for this.  This is for the ServiceServer, not Subscriber.

PR #301, opened 2016-02-04
Adds getParam() functions to swri_roscpp.

evenator:
These functions wrap NodeHandle::getParam(). If the parameter does
not exist, they emit an error message and return false.

### DIFF
--- a/swri_roscpp/include/swri_roscpp/parameters.h
+++ b/swri_roscpp/include/swri_roscpp/parameters.h
@@ -5,44 +5,100 @@
 
 namespace swri
 {
-static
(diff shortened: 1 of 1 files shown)


evenator:
The Travis build is broken due to unrelated issues, but this is the same as #300, so I'm going to go ahead and merge.

130s:
> The Travis build is broken due to unrelated issues, but this is the same as #300, so I'm going to go ahead and merge.

Sorry for the inconvenience and you're right. It should be fixed now. If you still see the similar issue please report at https://github.com/ros-industrial/industrial_ci/issues/26.

PR #306, opened 2016-03-14
Swri rospy

evenator:
A new package swri_rospy with a single-threaded callback queue and service wrapper decorator.


shaun-edwards:
@evenator, out of curiosity, did you find the current (multi-threaded) python implementation problematic?

pjreed:
We ran into an issue where we had a Python node that was both subscribing and publishing to topics, and a subscriber was modifying an object that the publisher was publishing; concurrency issues lead to the publisher publishing a bad message that would kill any node that tried to deserialize it.

This kind of thing isn't normally an issue with C++ nodes since they're really all single-threaded under the hood.

shaun-edwards:
There is an option to create multi-threaded C++ nodes with a multi-threaded spinner, but by default C++ is single threaded.  This discrepancy between roscpp and rospy has always bothered me.  Several people on our team have made the mistake of not writing thread-safe python code, so I think this is good thing to have.

PR #307, opened 2016-04-07
Fixed compile error when ros-indigo-opencv3 is installed

kimmat:
I was unable to compile the marti_common when ros-indigo-opencv3 was installed. This patch fixed the issue.


pjreed:
Good catch.  I've noticed that swri_image_util, swri_opencv_util, and swri_transform_util also all depend on OpenCV 2 but do not specify the version number; could you add those to your pull request?

kimmat:
I updated the CMakeFiles.txt. It would be great if you could merge it into your repository.

evenator:
LGTM. @pjreed, should we add a Travis test for this?

pjreed:
Yeah, probably.  The easiest way to test it would probably be to just manually install the ros-indigo-opencv3 packge before compling; it's very obvious when it fails.

PR #309, opened 2016-04-08
Fixed compile error when ros-indigo-opencv3 is installed (#307) (Jade)

pjreed:
Porting the changes in MR #307 to Jade.


pjreed:
I fixed other issues that were preventing Travis from compiling the jade-devel branch, so this should be all good now.

PR #330, opened 2016-05-08
Add util functions to calculate distances between route points. (indigo)

elliotjo:
This commit adds two utility functions to calculate the distances (in
terms of arc length) between route points.  One function calculates
the distance between two points, the other calculates the distance
between one point and many other points and should provide much better
performance for that common need.

### DIFF
--- a/swri_route_util/src/util.cpp
+++ b/swri_route_util/src/util.cpp
@@ -454,19 +454,6 @@ bool routeDistances(
   std::vector<double> arc_lengths;
   arc_lengths.resize(max_index-min_index+1);
 
-  // arc_lengths[0] = 0.0;
-  // for (size_t i = 1; i < arc_lengths.size(); i++) {
-  //   const tf::Vector3 pt1 = route.points[min_index+i].position();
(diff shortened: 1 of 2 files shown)


swri_route_util/src/util.cpp
```diff
@@ -373,4 +377,140 @@ bool interpolateRoutePosition(RoutePoint &dst,
...
+  // This is the index of the start point in the ROI.
+  const size_t roi_start_index = start_index - min_index;
+
+  // We calculate the arc length of each point in the ROI relative to
+  // the start point.  This vector covers the ROI (so it corresponds
+  // from min_index to max_index)
+  std::vector<double> arc_lengths;
+  arc_lengths.resize(max_index-min_index+1);
+
+  // arc_lengths[0] = 0.0;
```
malban: Is there a reason this commented out code is left in?

elliotjo: Good catch.  That was the straightforward version that I did as a reference for writing the actual version I ended up with.  Added commits to both merge requests to remove it.

PR #351, opened 2016-06-13
Fix kinetic build

evenator:
These changes allow marti_common to build for ROS Kinetic under Ubuntu Xenial. They don't break the build for ROS Jade, and they shouldn't change functionality at all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,17 +11,18 @@ notifications:
 env:
   global:
     - USE_DEB=true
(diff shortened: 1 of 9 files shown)


evenator:
Because Qt5 is incompatible with Qt4, I'm going to close this and recommend we make a new kinetic-devel branch based on it instead. It means we'll have to support 3 different branches (indigo, jade, and kinetic) at least until Jade EOLs in May 2017.

PR #352, opened 2016-06-15
Add a CMake macro to automatically create wrapper nodes

evenator:
Defines a CMake macro `swri_nodelet_add_node(NODELET_NODENAME NODELET_TYPE)` that automatically generates the c++ code for a node wrapper with node name `NODELET_NODENAME` that wraps the nodelet of type `NODELET_TYPE` and makes a CMake target to build the node.

### DIFF
--- a/swri_nodelet/CMakeLists.txt
+++ b/swri_nodelet/CMakeLists.txt
@@ -12,19 +12,41 @@ set(RUNTIME_DEPS ${BUILD_DEPS})
 find_package(catkin REQUIRED COMPONENTS ${BUILD_DEPS})
 catkin_package(
   CATKIN_DEPENDS ${RUNTIME_DEPS}
+  INCLUDE_DIRS include
+  CFG_EXTRAS swri_nodelet-extras.cmake
 )
 
-include_directories(${catkin_INCLUDE_DIRS})
+# nodelet.cpp.in must be copied before include(...) or clean builds will fail
+file(COPY nodelet.cpp.in
+   DESTINATION ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_SHARE_DESTINATION}
+)
+set(swri_nodelet_PREFIX ${CATKIN_DEVEL_PREFIX})  # Allows use of nodelet.cpp.in in swri_nodelet_add_node macro for this build
+include(${CATKIN_DEVEL_PREFIX}/share/${PROJECT_NAME}/cmake/swri_nodelet-extras.cmake)  # Allows use of swri_nodelet_add_node macro for this build
+
+include_directories(${catkin_INCLUDE_DIRS} include)
 
 add_library(swri_nodelet_test src/test_nodelet.cpp)
 target_link_libraries(swri_nodelet_test ${catkin_LIBRARIES})
+
+swri_nodelet_add_node(test_node swri_nodelet TestNodelet)
+target_link_libraries(test_node swri_nodelet_test)
 
 if(CATKIN_ENABLE_TESTING)
     find_package(rostest REQUIRED)
     add_rostest(test/test_manager.test DEPENDENCIES swri_nodelet_test)
     add_rostest(test/test_standalone.test DEPENDENCIES swri_nodelet_test)
+    add_rostest(test/test_node_wrapper.test DEPENDENCIES test_node)
 endif()
 
 install(PROGRAMS 
   nodes/nodelet
   DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
+
+install(FILES nodelet.cpp.in
+   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
+)
+
+install(DIRECTORY include/${PROJECT_NAME}/
+  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+  FILES_MATCHING PATTERN "*.h"
+)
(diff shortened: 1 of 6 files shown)


evenator:
The node wrapper is based on [this template](https://www.clearpathrobotics.com/assets/guides/ros/Nodelet%20Everything.html#and-now-code) from Clearpath.

evenator:
Example usage:

```
cmake_minimum_required(VERSION 2.8.3)
project(my_project)
find_package(catkin REQUIRED COMPONENTS 
    nodelet
    swri_nodelet
)

catkin_package(
  LIBRARIES ${PROJECT_NAME}_nodelets
  CATKIN_DEPENDS nodelet
)

include_directories(include)

## Declare nodelet library
add_library(${PROJECT_NAME}_nodelets
  src/nodelets/my_nodelet.cpp
)
target_link_libraries(${PROJECT_NAME}_nodelets ${catkin_LIBRARIES})

## Generate a node from the nodelet
swri_nodelet_add_node(my_node my_project/my_nodelet)
```

elliotjo:
I recommend a different node wrapper.  This is what I typically use:

namespace PACKAGE_NS {
boost::shared_ptr<nodelet::Nodelet> createMyNodelet();
}  // namespace PACKAGE_NS

int main(int argc, char **argv)
{
  ros::init(argc, argv, "map_registration");
  nodelet::M_string remappings;
  nodelet::V_string my_argv;  
  auto nodelet = PACKAGE_NS::createMyNodelet();
  nodelet->init(ros::this_node::getName(), remappings, my_argv);
  ros::spin();
  return 0;
}

It does have the downside that you have to define the createMyNodelet function in the cpp file containing the nodelet definition:

boost::shared_ptr<nodelet::Nodelet> createMyNodelet()
{
  return boost::make_shared<MyNodelet>();
}
}  // namespace PACKAGE_NS

but is has the benefit that it still works when your nodelet plugin XML file is wrong and more importantly, it catches missing function definitions at compile time.  Otherwise, the typical compilation succeeds but then the nodelet crashes at runtime with a missing symbol error.

evenator:
> it catches missing function definitions at compile time

That's a major plus in my book.

> It does have the downside that you have to define the createMyNodelet function in the cpp file containing the nodelet definition

I dislike that, because it means it doesn't work on a bog-standard nodelet.

> it has the benefit that it still works when your nodelet plugin XML file is wrong

I'd actually call that a downside, because it means you can be testing with the wrapped node and everything works, but then you switch to the nodelet and suddenly things are broken.

I think I might be able to find a way to synthesize these two methods together into something with the advantages of both...

evenator:
I can't think of a way to do this that doesn't require defining that factory function. With that in mind, I'm not sure which approach I favor. I definitely like the advantage of linking at compile time, but I would like this macro to be usable on arbitrary nodelets.

If we end up going with the factory function/pointer approach, I think it should be accompanied by a compile-time checker for the plugin XML file, but that would probably be a good thing to have anyway.

Alternatively, a third approach would be to make a macro for creating a nodelet *library*. You give it a list of c++ files and it autogenerates a c++ file with the factory functions,  generates the library target, then generates the c++ and targets for all of the wrapper nodes. I like that approach, but it might be inappropriate to put that much magic in a build system.

shaun-edwards:
Tagging @akgoins and @Jmeyer1292 because I think we would find this useful.

evenator:
I just came up with another idea, so here are the 4 approaches I've thought of

1. The approach used in this PR. Advantages are that it can be used on any nodelet, and that it will succeed or fail in the same way as a nodelet. It has the disadvantage that, like nodelets, it won't generate compile-time linker errors.
2. @elliotjo's approach. This approach has the advantage of generating compile-time linker errors. It has the disadvantages that it requires you to define a factory function and that, since it doesn't use dynamic loading, it doesn't expose dynamic loading errors.
3. Use elliotjo's approach, but provide a C++ macro to generate the factory function. Instead of `PLUGINLIB_EXPORT_CLASS(example_pkg::MyNodeletClass, nodelet::Nodelet)`, declare the nodelet with `SWRI_NODELET_EXPORT_CLASS(example_pkg::MyNodeletClass, nodelet::Nodelet)`, which would wrap `PLUGINLIB_EXPORT_CLASS()` and also generate the factory function.
4. Provide a CMake macro for nodelet library generation. You give it a list of c++ files and it autogenerates a c++ file with the factory functions, generates the library target, then generates the c++ and targets for all of the wrapper nodes. The primary disadvantage of this is that it's very complex and puts a lot of magic in the build system.

Of those options, I like 3 the best, followed by 1, then 4. 2 is a subset of 3 since the developer could  choose not to use the macro and just write the factory themselves. I think options 2-4 would benefit from some checks to ensure that dynamic loading is set up correctly and working.

I'm going to start implementing 3, but if anyone has anymore input, I'd appreciate it.

evenator:
I've implement number 3 above. In the nodelet's c++ file use the following code for the class list macro:

    #include <swri_nodelet/class_list_macros.h>
    SWRI_NODELET_EXPORT_CLASS(namespace, ClassName);

This exports the nodelet class as usual and also creates a factory function `boost::shared_ptr<nodelet::Nodelet> createClassName();` for the nodelet.

In the CMakeLists, to create a wrapper node, use the CMake macro:

    swri_nodelet_add_node(node_name namespace ClassName)
    target_link_libraries(node_name library_name)

This creates a node executable with the specified node_name using @elliotjo's wrapper.

One feature that's still missing is a test that the node is dynamically loadable. Since the node wrapper uses a shared pointer to load the nodelet, there's no coverage for dynamic loading, so if the nodelet isn't specified in the plugin XML file, it would work in the node wrapper but not as a nodelet.

I've updated the tests to cover the new macros, and all tests are passing. I think this request is ready for review and merge.

elliotjo:
Looks great, nice work!

PR #355, opened 2016-06-27
Fix projectOntoRoute for points past the end of the route. (indigo)

elliotjo:
This fixes projectOntoRoute to return a normalized route coordinate
when the point is past the end of the route.

Needs testing on vehicle.


malban:
Tested on vehicle.

PR #357, opened 2016-06-29
Add sru::projectOntoRouteWindow

elliotjo:
This is a utility function to project a point onto a window of the
route.


malban:
Tested on vehicle.

PR #359, opened 2016-06-30
Add native support for the "stop_point" property in route points.

elliotjo:
This commit adds native support for the "stop_point" property for
swri_route_util::RoutePoint.  This property defaults to false when it
is not explicitly set to true in the route point.


malban:
Tested on vehicle

PR #385, opened 2016-10-11
Adds the ability to warn when unused parameters are set in a namespace.

evenator:
A common error when using unfamiliar ROS nodes is to accidentally set parameters by the wrong name. This feature allows the node author to output a WARNING for any unused parameters.

See the param_test node for an example.


evenator:
Yeah, I was following convention there. I can rewrite that as a unit test.

swri_roscpp/include/swri_roscpp/parameters.h
```diff
@@ -1,15 +1,30 @@
...
+
+#include <ros/console.h>
 #include <ros/node_handle.h>
 
 namespace swri
 {
+
+  /// This set stores all of the parameters that have been got with this library
+  /// It is used for the getUnusedParamKeys and WarnUnusedParams functions
+  std::set<std::string> _used_params;
```
elliotjo: This either needs to be declared static or extern.  extern makes more sense to work well with nodes that are spread across several files, but means swri_roscpp will have to produce an actual library that is linked into dependent code.  I'd also request that the prefix be changed from _ to g_ but that's a personal preference.

swri_roscpp/include/swri_roscpp/parameters.h
```diff
@@ -126,8 +153,65 @@ namespace swri
...
+	std::vector<std::string> unused_params(all_params.size());
+	boost::function<bool(std::string)> inNamespace = boost::bind(isPrefixOf, n_namespace, _1);
+	std::vector<std::string>::const_iterator first_param_in_namespace = find_if(all_params.begin(), all_params.end(), inNamespace);
+	std::vector<std::string>::const_iterator after_last_param_in_namespace = find_if(all_params.begin(), all_params.end(), std::not1(inNamespace));
+	std::vector<std::string>::iterator last_unused_param = std::set_difference(first_param_in_namespace, after_last_param_in_namespace, _used_params.begin(), _used_params.end(), unused_params.begin(), prefixLessThan);
+	unused_params.resize(last_unused_param - unused_params.begin());
+	return unused_params;
+  }
+
+  void warnUnusedParams(ros::NodeHandle const& n = ros::NodeHandle("~"))
```
elliotjo: Let's get rid of the default value.  It won't work with nodelets and we'll end up with lots of "we'll fix that later" problems.

swri_roscpp/include/swri_roscpp/parameters.h
```diff
@@ -126,8 +153,65 @@ namespace swri
...
+	n.getParamNames(all_params);
+	std::sort(all_params.begin(), all_params.end());
+	std::vector<std::string> unused_params(all_params.size());
+	boost::function<bool(std::string)> inNamespace = boost::bind(isPrefixOf, n_namespace, _1);
+	std::vector<std::string>::const_iterator first_param_in_namespace = find_if(all_params.begin(), all_params.end(), inNamespace);
+	std::vector<std::string>::const_iterator after_last_param_in_namespace = find_if(all_params.begin(), all_params.end(), std::not1(inNamespace));
+	std::vector<std::string>::iterator last_unused_param = std::set_difference(first_param_in_namespace, after_last_param_in_namespace, _used_params.begin(), _used_params.end(), unused_params.begin(), prefixLessThan);
+	unused_params.resize(last_unused_param - unused_params.begin());
+	return unused_params;
+  }
```
elliotjo: How about we go simpler here with something like:

std::vector<std::string> result;
for (blahblah it = all_params.begin(); it != all_params.end(); ++it) {
  if (*it.compare(0, n_namespace.size(), n) == 0 && used_params.count(*it) == 0) {
    result.push_back(*it);
  }
}

swri_roscpp/include/swri_roscpp/parameters.h
evenator: Sounds reasonable.

swri_roscpp/include/swri_roscpp/parameters.h
evenator: I'll preface this by saying that I wrote this partly as an exercise in playing with the std algorithm library.

I agree that std algorithm code isn't easy on the eyes. However, the simpler algorithm would false-positive on members of arrays and structs. Admittedly, we don't yet have getters for those types, but that was the thinking. This filter is also faster than the simpler algorithm, which may or may not matter, depending on how many parameters are used by the node and are in the system.

Overall, I agree this needs a second pass to make it simpler, so I'll look into doing it.

evenator:
I've added an automated unit test for `getUnusedParameterKeys()`, per @pjreed's request. I've also removed the default value for the NodeHandle argument to `warnUnusedParameters()`, per @elliotjo's request.

Regarding simplifying the algorithm, I've concluded that the algorithm I wrote originally is correct, but I agree that it was inscrutable as written. I've attempted to alleviate that problem by adding more comments to describe what's going on. Is that an acceptable improvement?

evenator:
@pjreed Can you update your review? I added a unit test that can run in CI.

@pjreed, @elliotjo, @malban Is this merge-able now? Jason and Neal are interested in using it.

malban:
Fine with me

PR #387, opened 2016-10-14
Install nodelet_plugins.xml

evenator:


elliotjo:
Do we really need to install this file when it's only for unit testing?

evenator:
That was my original stance as well, but if you have it listed in the package.xml and it isn't present, you get an error message every time you load a nodelet.

PR #395, opened 2016-11-30
Migrated OpenCV to 3.1 (default in Kinetic)

bdholt1:
I made the changes required to compile and run mapviz on kinetic.  This is one of them.

### DIFF
--- a/swri_transform_util/package.xml
+++ b/swri_transform_util/package.xml
@@ -19,6 +19,7 @@
   <depend>geographic_msgs</depend>
   <depend>geometry_msgs</depend>
   <depend>libgeos++-dev</depend>
+  <depend>opencv3</depend>
   <depend>pluginlib</depend>
   <depend>proj</depend>
   <depend>roscpp</depend>
--- a/swri_transform_util/package.xml
+++ b/swri_transform_util/package.xml
@@ -13,13 +13,12 @@
   <url>https://github.com/swri-robotics/marti_common</url>
   
   <buildtool_depend>catkin</buildtool_depend>
-  
+
   <depend>boost</depend>
   <depend>diagnostic_msgs</depend>
   <depend>geographic_msgs</depend>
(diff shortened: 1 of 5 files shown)


evenator:
Relevant:

* [Kinetic Migration Guide: OpenCV](http://wiki.ros.org/kinetic/Migration#OpenCV)
* [ROS Wiki: OpenCV3](http://wiki.ros.org/opencv3)

swri_transform_util/package.xml
```diff
@@ -13,13 +13,12 @@
...
   
   <buildtool_depend>catkin</buildtool_depend>
-  
+
   <depend>boost</depend>
   <depend>diagnostic_msgs</depend>
   <depend>geographic_msgs</depend>
   <depend>geometry_msgs</depend>
   <depend>libgeos++-dev</depend>
-  <depend>libopencv-dev</depend>
```
pjreed: If we're not using the system-provided OpenCV 2 anymore, it makes sense to remove this, but would it make sense to add a rosdep key for the ROS-provided OpenCV 3?

bdholt1: Based on http://wiki.ros.org/opencv3 is seems that the key is opencv3, so I'll add that back.

evenator: See http://wiki.ros.org/opencv3#package.xml

bdholt1: Sorry, I was too quick on the trigger, I should have read the whole document first.

bdholt1:
Is there anything else that need to be done before this can be merged?

pjreed:
This looks good to me, although I'd like to get somebody else to test it out first (I don't have a Kinetic environment right at the moment), and it couldn't hurt to get @malban's approval.

malban:
Looks good to me.  We haven't tested it since we don't use kinetic yet, but I'll go ahead and merge it anyway.

PR #396, opened 2016-12-06
Fix issue with contrast stretching when a grid cell is completely masked out. (indigo)

malban:
Fix issue with contrast stretching when a grid cell is completely masked out.


kriskozak:
Looks good to me.

PR #404, opened 2017-01-13
Fix include_directories order in swri_roscpp

evenator:
Include package-local header files before catkin header files so that source is always built against source header files, even if the same package exists elsewhere in the catkin tree.


pjreed:
:+1:

PR #407, opened 2017-01-19
Check for error from ioctl in serial_port

### LINKED ISSUE #406: swri_serial_util segfaults on error
Line 307 of swri_serial_util/src/serial_port.cpp gets the number of available bytes using `iotctl( .. )`. On failure, this outputs -1 bytes to read. This is subsequently static cast to `size_t`, yielding 18446744071897977696 bytes to read. The program then tries to allocate that many bytes on line 312 and runs out of memory.

evenator:
Fixes #406

### DIFF
--- a/swri_serial_util/src/serial_port.cpp
+++ b/swri_serial_util/src/serial_port.cpp
@@ -305,6 +305,11 @@ namespace swri_serial_util
     {
       int bytes;
       ioctl(fd_, FIONREAD, &bytes);
+      if (bytes < 0)
+      {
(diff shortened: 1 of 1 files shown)


malban:
Can you squash this into 1 commit?

elliotjo:
I recommend adding more information to the error message to help figure out why the call fails:

error_msg_ = std::string("Error getting number of available bytes from serial port (") + strerror(errno) + ")";

evenator:
@malban I was planning to squash on merge.
@elliotjo  Yeah, I can make that change.

evenator:
Updated with @elliotjo's suggestion and squashed per @malban's request

PR #410, opened 2017-02-15
Include boost/make_shared.hpp

evenator:
This header is necessary in the unlikely event that the using cpp file hasn't included boost/make_shared.hpp anywhere else. Also, it's good practice to include what you use.

### DIFF
--- a/swri_nodelet/include/swri_nodelet/class_list_macros.h
+++ b/swri_nodelet/include/swri_nodelet/class_list_macros.h
@@ -30,6 +30,7 @@
 #ifndef SWRI_NODELET_CLASS_LIST_MACROS_H_
 #define SWRI_NODELET_CLASS_LIST_MACROS_H_
 
+#include <boost/make_shared.hpp>
 #include <pluginlib/class_list_macros.h>
 
 /*


elliotjo:
LGTM do we need to add this to jade/kinetic too?

evenator:
Yeah, it should be added for all three. Should I make PRs, or would you rather just cherry-pick after the merge?

On another note, the failed builds seem to be a travis hiccup--swri_yaml_util failing because of a read-only disk. I'll rerun them; hopefully they'll pass.

evenator:
Still failing. I was curious how this could be happening, so I reran a previously passing build, [262](https://travis-ci.org/swri-robotics/marti_common/builds/193541250), which is the head of indigo-devel, previously passed, and is known to build on multiple developer machines. It failed.

My best guess is that something has changed in Industrial CI since that build originally passed 28 days ago that no longer allows it to pass. @130s might be interested.

evenator:
Ok. I've nailed down the problem. I'll make a separate issue for it.

evenator:
The push test is failing because the branch hasn't been rebased onto the changes from #412 , but the PR test is passing (the PR test merges before testing). This should be good to merge.

PR #413, opened 2017-03-02
Dynamic reconfigurable transform publisher (Indigo)

malban:
Dynamic reconfigurable transform publisher

### DIFF
--- a/swri_transform_util/CMakeLists.txt
+++ b/swri_transform_util/CMakeLists.txt
@@ -1,41 +1,51 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(swri_transform_util)
 
-find_package(catkin REQUIRED COMPONENTS 
-  roscpp 
-  rospy
-  tf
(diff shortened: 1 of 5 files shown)


elliotjo:
FYI, you don't need to use a mutex for dynamic reconfig unless you're using multithreaded/asynchronous spinners.   The dynamic reconfig server works in the standard ROS event framework.

I'd also recommend adding a guard against initialization in the reconfigure callback.  This allows your ROS parameter default values to be the actual default values instead of the values defined in the cfg file:

void ConfigCallback(DynamicPublisherConfig &config, uint32_t level)
{
  boost::recursive_mutex::scoped_lock lock(config_mutex_);
  if (~level == 0) {
    // Initialize dynamic reconfig server with our loaded values at start up.
    config = config_;
  } else {
    config_ = config;
  }
}

malban:
@elliotjo Updated with your suggestions

PR #423, opened 2017-03-22
Support higher serial baud rates (Indigo)

pjreed:
The system header `/usr/include/asm-generic/termbits.h` has constants for
supporting baud rates up to 4000000, but swri_serial_util only allows up
to 230400.  Some devices support these higher rates and there's no reason
to not support them, so this adds support for them.


evenator:
LGTM

PR #426, opened 2017-04-06
Image blending

danthony06:
This adds a node for blending one image onto another.


pjreed:
Looks good to me, could you also cherry-pick this and make merge requests for `jade-devel` and `kinetic-devel`?

PR #439, opened 2017-04-25
Enable blending with transparency mask

### LINKED ISSUE #440: Add transparency option to image blending utility
Frequently, when we say we want to blend images, we really just want to blend portions of an image while leaving the remainder unchanged. The most common instance is when the top image is a black-and-white (or other two-color) mask-like image, and we wish to blend only one piece of it.

The image blender should take an optional mask color parameter to be made transparent in the top image.

jatowler:
Instead of blending an entire image, mask out portions that are a
particular color (e.g., for a black-and-white top image, only blend the
white portions, so the black portions don't darken the entire base image).

Adds three parameters to the `swri_image_util` `blend_images_nodelet` (mask
red, green, and blue components).

Fixes #440

### DIFF
--- a/swri_image_util/src/blend_images_util.cpp
+++ b/swri_image_util/src/blend_images_util.cpp
@@ -77,8 +77,6 @@ namespace swri_image_util
       const cv::Scalar mask_color,
       cv::Mat& dest_image)
   {
-    const bool debug = false;
-
     // All images must have the same shape. Return without modifying anything
     // if this is not the case
     if ((base_image.rows != top_image.rows)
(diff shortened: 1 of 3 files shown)


swri_image_util/src/blend_images_util.cpp
```diff
@@ -67,4 +69,91 @@ namespace swri_image_util
...
   }
+
+  void blendImages(
+      const cv::Mat& base_image,
+      const cv::Mat& top_image,
+      const double alpha,
+      const cv::Scalar mask_color,
+      cv::Mat& dest_image)
+  {
+    const bool debug = false;
```
pjreed: If this debugging code is something that's still useful to have in here, I'd rather this be a parameter so it can be turned on or off at runtime.  If it's not still useful, it should probably be removed.

jatowler: It was just developmental. I've removed it.

pjreed:
Looks good to me.  Could you also cherry-pick this into `jade-devel` and `kinetic-devel`?

PR #443, opened 2017-05-09
Merge together the indigo, jade, and kinetic branches

pjreed:
Maintaining three different branches for different ROS distributions is a pain and also unnecessary, so I've merged all of them together.  This branch builds and all tests pass on indigo, jade, kinetic, and lunar.  Don't look too hard at that commit list, it's ugly.

This PR targets the `indigo-devel` branch, but after it is merged, we should rename that branch to `master`, lock down all of the distro-specific branches, and then release a new version (0.3.0) to all distros.

Not very much needed to be changed to make everything compatible, but some notable things are:
1. Packages that use OpenCV (`swri_geometry_util`, `swri_image_util`, `swri_opencv_util`, and `swri_transform_util`) have a rosdep on `cv_bridge` and don't specify a version when calling `find_package` in their `CMakeLists.txt`; this means they'll use OpenCV 2 on Indigo and Jade and OpenCV 3 on Kinetic and Lunar.
2. `swri_image_util` depends on Qt 5.
3. Nodes and utilities in `swri_transform_util` previously used the `gps_common/GPSFix` message on Indigo and `sensor_msgs/NavSatFix` (or Pose, or GeoPose, or whatever else was appropriate) on Jade and later.  Now they will all handle all of them.  This is a little ugly in `initialize_origin.py` because it was basically completely rewritten for NavSatFix support.  The right thing to do would be to rewrite it again to handle every case gracefully, but that's a bit beyond the scope of this PR, so instead it just detects whether your ${ROS_DISTRO} is "indigo" or not and changes behavior appropriately.


evenator:
I recommend we soak test this in SwRI-owned vehicles for a while before merging. A major change to marti_common affects a lot of projects. We should also verify that the kinetic packages still work as expected. I'm not aware of any projects using Jade, so I'm not too concerned about that.

swri_yaml_util/CMakeLists.txt
```diff
@@ -28,10 +28,10 @@ configure_file(version.h.in ${CATKIN_DEVEL_PREFIX}/${CATKIN_PACKAGE_INCLUDE_DEST
 find_package(Boost REQUIRED)
 
 include_directories(include
-  ${CATKIN_DEVEL_PREFIX}/${CATKIN_GLOBAL_INCLUDE_DESTINATION}
-  ${catkin_INCLUDE_DIRS}
   ${Boost_INCLUDE_DIRS}
+  ${CATKIN_DEVEL_PREFIX}/${CATKIN_GLOBAL_INCLUDE_DESTINATION}
   ${YamlCpp_INCLUDE_DIR}
+  ${catkin_INCLUDE_DIRS}
```
evenator: As a rule, I prefer to include in the order:

1. package-local
2. catkin
3. system

I do this on the principal that a catkin workspace overlay should take precedence over system packages.

pjreed: Good point.  Both `swri_yaml_util` and `swri_transform_util` didn't have `${catkin_INCLUDE_DIRS}` second, so I fixed them.

pjreed:
I found one hitch on Indigo; Mapviz uses `swri_image_util`, which means that if `swri_image_util` is switched to Qt 5, Mapviz would need to use Qt 5, and anything that provides a plugin for Mapviz would also need to switch.  In addition, anything that links in OpenCV's `highgui` module would also require Qt 4.

It's easy to make `swri_image_util` switch which version of Qt  it links it, but because there's no way in `package.xml` to make dependencies conditional, that means a cross-distro compatible version has to say it depends on both Qt 4 and 5.  It doesn't hurt anything, but it does mean you'd end up installing a version of Qt you don't need.

swri_image_util/package.xml
```diff
@@ -20,6 +20,8 @@
   <depend>geometry_msgs</depend>
   <depend>image_geometry</depend>
   <depend>image_transport</depend>
+  <depend>libqt4-opengl-dev</depend>
```
evenator: Will having both qt4 and qt5 as deps cause a conflict?

pjreed: They don't conflict at all, and in fact work fine when installed side-by-side.  It's mostly just annoying that rosdep is guaranteed to install a version of Qt that you may not be using at all.

pjreed:
Easier solution: remove Qt entirely.  `swri_image_util` was only using its QPolygonF class to find the intersection of two polygons, but `swri_geometry_util` can do that even more easily, so Qt is entirely unnecessary in marti_common.  All unit tests pass.

evenator:
I'm going to set a target to get the necessary tests in on this next week, with the goal to merge by June 2.

elliotjo:
What tests are needed? I would love to get this merged as soon as possible.

evenator:
This totally got pushed off my plate. Sorry about that. @pjreed tested it in Kinetic. I don't care about Jade, but it needs to be tested for Indigo. I'd hate to see a *major government client* 😉 have problems.

nick-alton:
Tested on vehicle, there seems to be an issue with initialize_origin.py in swri_transform_util. When it gets to line 118 it complains that global name '_gps_fix' is not defined.

evenator:
Ok, @nick-alton, try it again now.

Issue #281, opened 2015-12-11
Improve Doxygen API Documentation


evenator:
The documentation for these packages is sparse (where it exists at all) and sometimes incorrect. For example, see http://docs.ros.org/api/swri_transform_util/html/.

PR #454, opened 2017-06-19
Document swri tranform util

evenator:
This takes a chunk out of the effort to document `swri_transform` util. There's still a lot to be done, but this improves the `rosdoc` configuration and adds some more doxygen documentation for that package.

Refs (but does not close) #281

### DIFF
--- a/swri_transform_util/include/swri_transform_util/earth_constants.h
+++ b/swri_transform_util/include/swri_transform_util/earth_constants.h
@@ -37,6 +37,9 @@ namespace swri_transform_util
    */
   static const double _earth_equator_radius = 6378137.0;
 
+  /**
+   * Earth mean radius
+   */
   static const double _earth_mean_radius = 6371009.0;
 
   /**
--- a/swri_transform_util/include/swri_transform_util/local_xy_util.h
+++ b/swri_transform_util/include/swri_transform_util/local_xy_util.h
@@ -80,14 +80,25 @@ namespace swri_transform_util
       double& longitude);
 
   /**
-   * Utility class for converting between WGS84 lat/lon and an ortho-rectified
+   * @brief Utility class for converting between WGS84 lat/lon and an ortho-rectified
    * LocalXY coordinate system.
+   *
+   * To use this class, first construct it with a reference origin. The
+   * reference origin should be a latitude, longitude, angle, and altitude in
+   * WGS84 coordinates. Once initialized, a LocalXyWgs84Util can be used to
+   * convert WGS84 coordinates to and from an ortho-rectified frame with its
+   * origin at the reference origin. Because the earth is spherical, the error
+   * in the ortho-rectified frame will accumulate as the distance from the
+   * reference origin increases. For this reason, the reference origin should
+   * be chosen to be close to the region of interest (<10 km).
    */
   class LocalXyWgs84Util
   {
   public:
     /**
-     * Constructor.
+     * @brief Initializing constructor
+     *
+     * This constructor creates and initializes a LocalXyWgs84Util.
      *
      * @param[in] reference_latitude   Reference latitude in degrees.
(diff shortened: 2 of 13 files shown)


kriskozak:
One comment: I just opened one of the files (swri_transform_util/local_xy_util.h) randomly and saw this:

> To use this class, first construct it with a reference origin. The reference origin should be a latitude, longitude, angle, and altitude in WGS84 coordinates.

I would recommend discouraging the use of non-zero angles in the documentation (like adding a parenthetical "usually angle should be zero except for special cases"). I would like to remove the angle parameter altogether... but that ship has sailed, and I just noticed that angle is the third of four parameters (the fourth being altitude, which should usually be set to a valid altitude if it is known). Too bad.

swri_transform_util/include/swri_transform_util/local_xy_util.h
```diff
@@ -80,14 +80,25 @@ namespace swri_transform_util
       double& longitude);
 
   /**
-   * Utility class for converting between WGS84 lat/lon and an ortho-rectified
+   * @brief Utility class for converting between WGS84 lat/lon and an ortho-rectified
```
malban: Not that it matters much, but I think the @brief tag here and elsewhere is redundant.  The first paragraph is automatically considered the brief description by Doxygen  and I don't think adding the tag adds anything for human readability.

swri_transform_util/include/swri_transform_util/earth_constants.h
```diff
@@ -37,6 +37,9 @@ namespace swri_transform_util
    */
   static const double _earth_equator_radius = 6378137.0;
 
+  /**
+   * Earth mean radius
```
pjreed: A comment that just echoes the name of the variable is not very useful.  Maybe also say what units it's in?

swri_transform_util/include/swri_transform_util/local_xy_util.h
pjreed: Yep, `@brief` is pointless unless you want to have multiple `@brief` sections that are joined together in the generated documentation.  (and if you do that you've probably missed the point of `@brief`)

evenator: `@brief` is necessary unless `autobrief` is turned on, and in `rosdoc` it isn't. You can see [here](http://docs.ros.org/indigo/api/swri_transform_util/html/classswri__transform__util_1_1UtmUtil.html) that none of the methods of `UtmUtil` have briefs, even though the [source code](https://github.com/swri-robotics/marti_common/blob/master/swri_transform_util/include/swri_transform_util/utm_util.h) has documentation.

swri_transform_util/include/swri_transform_util/earth_constants.h
evenator: I agree. What units is it in? I would assume meters, but I didn't write the constant.

evenator:
Rebased and reopened as #456

swri_transform_util/include/swri_transform_util/earth_constants.h
pjreed: Yep, I double-checked and it's meters.

swri_transform_util/include/swri_transform_util/local_xy_util.h
pjreed: Oh, so it is.  I wonder what it'd take to enable autobrief for a project...

evenator: There's a yaml config file for rosdoc. It might be possible to enable it there, but I don't know what options it supports.

swri_transform_util/include/swri_transform_util/earth_constants.h
evenator: Resolved.

PR #456, opened 2017-06-22
Document swri_transform_util

evenator:
Reopening #454, which was based on `indigo-devel`. This one is based on the new `master` branch.


evenator:
I addressed the three issues with #454:

- @kriskozak's comment about using 0 angle
- @malban and @pjreed's comments about enabling autobrief
- @pjreed's comment about the Earth mean radius variable

Issue #452, opened 2017-06-15
[swri_transform_util] initialize_origin.py Needs to Be Rewritten


evenator:
The already unwieldy code for this node is now completely unmanageable after the changes in #443. It should be completely rewritten from the ground up.

Requirements (IMHO):

- Must not break existing API in ROS Indigo or Kinetic. This means it must accept either `GPSFix` or `NavSatFix` messages for automatic origin initialization. 
- The node name `initialize_origin.py` must still work (although I would encourage making that a thin wrapper around a new node name).
- Simplify the code and reorganize it. It might also be worthwhile to design it as a class that could be incorporated into a larger node (this feature has been requested for Mapviz more than once.)

PR #457, opened 2017-06-26
Initialize origin tests

evenator:
Add some tests for `initialize_origin.py` in preparation for #452. These tests are far from exhaustive, but I think they're sufficient to test parameter API compatibility and interoperability with the CPP origin utilities.

### DIFF
--- a/swri_transform_util/test/test_initialize_origin.py
+++ b/swri_transform_util/test/test_initialize_origin.py
@@ -77,7 +77,7 @@ class TestInitializeOrigin(unittest.TestCase):
             latitude = msg.latitude
             longitude = msg.longitude
             altitude = msg.altitude
-            self.assertEqual(heading, swri['heading'])
+            self.assertEqual(msg.track, swri['heading'])
         else:  # msg._type == GeoPose._type:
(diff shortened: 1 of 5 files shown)


swri_transform_util/test/test_initialize_origin.py
```diff
@@ -0,0 +1,144 @@
...
+            latitude = msg.position.latitude
+            longitude = msg.position.longitude
+            altitude = msg.position.altitude
+            quaternion = (msg.pose.orientation.x,
+                           msg.pose.orientation.x,
+                           msg.pose.orientation.x,
+                           msg.pose.orientation.x)
+            euler = tf.transformations.euler_from_quaternion(quaternion)
+            yaw = euler[2]
+            self.assertEqual(yaw, 0)
```
pjreed: Not sure exactly what math `euler_from_quaternion` is doing internally, but should this (and other tests like it) be `assertAlmostEqual` instead in case there are some floating point issues?

evenator: Yes, they should.

PR #461, opened 2017-07-20
[swri_image_util] Add a crosshairs nodelet to image_util

evenator:
This new node[let] `swri_image_util/crosshairs` draws a crosshairs in the center of an image and publishes the drawn-upon image for display in a tool such as `image_view`. This is useful for aligning things in the real world with the center of the image.

The majority of the changes to nodelet_plugins.xml are from alphabetizing the nodelets.

### DIFF
--- a/swri_image_util/src/nodelets/crosshairs_nodelet.cpp
+++ b/swri_image_util/src/nodelets/crosshairs_nodelet.cpp
@@ -66,11 +66,18 @@ namespace swri_image_util
     void ImageCallback(const sensor_msgs::ImageConstPtr& image)
     {
       cv_bridge::CvImagePtr cv_image = cv_bridge::toCvCopy(image);
+      // Get image dimensions
       const int h = cv_image->image.rows;
       const int w = cv_image->image.cols;
+      // Define line properties
       const cv::Scalar black(0, 0, 0);
-      cv::line(cv_image->image, cv::Point(0, w/2), cv::Point(h-1, w/2), black, 3, 8);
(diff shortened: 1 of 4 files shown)


swri_image_util/src/nodelets/crosshairs_nodelet.cpp
```diff
@@ -0,0 +1,85 @@
...
+      image_sub_ = it.subscribe("image", 1, &CrosshairsNodelet::ImageCallback, this);
+    }
+
+    void ImageCallback(const sensor_msgs::ImageConstPtr& image)
+    {
+      cv_bridge::CvImagePtr cv_image = cv_bridge::toCvCopy(image);
+      const int h = cv_image->image.rows;
+      const int w = cv_image->image.cols;
+      const cv::Scalar black(0, 0, 0);
+      cv::line(cv_image->image, cv::Point(0, w/2), cv::Point(h-1, w/2), black, 3, 8);
```
jatowler: Do these 3 and 8 mean something? I'm not sufficiently familiar with OpenCV to know the `cv::line()` syntax. You might prefer

```
const int thickness = 3;
const int line_type = 8; // 8-connected line
```

for clarity.

jatowler:
I made one suggestion for clarity; other than that, LGTM.

swri_image_util/src/nodelets/crosshairs_nodelet.cpp
evenator: That's a good point.

Issue #370, opened 2016-09-18
swri_transform_util::Transform has Unimplemented Inverse() Method


evenator:
The Inverse() method is declared in transform.h, but never implemented in transform.cpp

malban:
This was fixed in #464

PR #464, opened 2017-07-25
Add inverse transform implementation to transforms.

malban:
Add inverse transform implementation to transforms.

Unit tests have been updated to test this method.


evenator:
This should close #370

malban:
There is apparently an issue with me changing the library name of the plugins in this commit.

malban:
Should be fixed now.

PR #469, opened 2017-07-28
Integrate transformers as static classes instead of plug-ins.

malban:
Integrate transformers as static classes instead of plug-ins.

This should fix the weird test failures.


evenator:
LGTM. When it passes CI, I'll merge.

evenator:
Oddly, this doesn't fix the flaky test.

PR #481, opened 2017-08-26
Add CI lint, better CI organization

evenator:
- Use Travis's new Stages and Jobs instead of Matrix
- Add a lint job to run catkin_lint on the repository
- Prioritize required jobs over jobs that are allowed to fail


evenator:
Note that the catkin lint check is currently failing, but I've marked it as allowed to fail.

PR #483, opened 2017-08-29
Add OptionalDiagnosedPublisher class

pjreed:
This is identical to diagnostic_updater::DiagnosedPublisher except that it can be enabled or disabled; when disabled, it will publish an "OK" diagnostic indicating that it has been disabled.

This is useful for when you have a DiagnosedPublisher for a topic in order to monitor its publish rate but it is acceptable for that topic to be turned off so that it's not publishing anything.

### DIFF
--- a/swri_roscpp/include/swri_roscpp/optional_diagnosed_publisher.h
+++ b/swri_roscpp/include/swri_roscpp/optional_diagnosed_publisher.h
@@ -1,0 +1,97 @@
+// *****************************************************************************
+//
+// Copyright (c) 2017, Southwest Research Institute® (SwRI®)
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


pjreed:
If you want to test this, you can do so against the `6-diagnostics-show-errors-for-unpublished-topics` branch on our internal `microstrain_3dm_gx4` repository.

evenator:
Isn't the `override` specifier C++11 only?

pjreed:
Indeed it is.  Since this is a header-only class, it doesn't actually matter for the compilation of this package, but I suppose it would make sense to remove that since nothing else in this class uses C++11.

evenator:
Yeah, I'd like to avoid `swri_cpp` requiring C++11, especially since it causes problems with PCL.

pjreed:
The status on this branch seems to have timed out for some reason, but the build was successful: https://travis-ci.org/swri-robotics/marti_common/builds/269656297

PR #485, opened 2017-09-20
Add support for vehicle_width_override property on route

matt-attack:


swri_route_util/src/route_speeds.cpp
```diff
@@ -355,10 +355,24 @@ void speedsForObstacles(
       }
       continue;
     }
+    
+    // Use lane width as the car radius when smaller
```
malban: Please elaborate the comment to describe the purpose of this override and replace 'lane width' with the name of the propery

swri_route_util/src/route_speeds.cpp
```diff
@@ -355,10 +355,27 @@ void speedsForObstacles(
...
+    double veh_r = car_r;
+    if (point.hasProperty("vehicle_width_override"))
+    {
+      ROS_DEBUG("Speeds for obstacle found vehicle_width_override property");
+      double width = point.getTypedProperty<double>("vehicle_width_override");
+
+      // Pick the smaller of the radii
+      if (veh_r >= width/2.0)
+      {
+        veh_r = width/2.0;
```
kriskozak: It probably makes sense to issue a warning here if the vehicle width is actually overridden. Doing so won't change the behavior, but it will make it more obvious if it causes problems. Given Marc's earlier comment about spamming the console, a `ROS_WARN_THROTTLE(1.0, ...)` is probably appropriate.

malban: Good point.

matt-attack: Added

PR #486, opened 2017-09-20
Add bounds checking to extractSubroute.

malban:
Add bounds checking to extractSubroute to handle start and end positions being out of order.

### DIFF
--- a/swri_route_util/src/util.cpp
+++ b/swri_route_util/src/util.cpp
@@ -711,6 +711,8 @@ bool extractSubroute(
 
   if (end_index <= start_index)
   {
+    sub_route.points.clear();
+    sub_route.rebuildPointIndex();
     return true;
   }
 
--- a/swri_route_util/src/util.cpp
+++ b/swri_route_util/src/util.cpp
@@ -709,6 +709,11 @@ bool extractSubroute(
   end_index++;
   end_index = std::min(end_index, route.points.size());
 
+  if (end_index <= start_index)
+  {
+    return true;
+  }
+
   sub_route.points.reserve(end_index - start_index);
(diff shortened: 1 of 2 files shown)


swri_route_util/src/util.cpp
```diff
@@ -709,6 +709,11 @@ bool extractSubroute(
   end_index++;
   end_index = std::min(end_index, route.points.size());
 
+  if (end_index <= start_index)
+  {
+    return true;
+  }
+
```
kriskozak: Did you ask me to review this just to see if I was paying attention?  

I'm not sure I follow all of the logic here 100%, but or this part, does this actually constitute a success?  Or did you mean to return false. If it's fine, should you have cleared sub_route.points `sub_route.points.clear()` before returning here or up at the top of this function?

elliotjo:
I'd say either clear it right after the test and return true, or return
false and don't touch sub_route.   The first one seems like the most
convenient to me (sort of in line with python slices).

swri_route_util/src/util.cpp
malban: I think it should return true.  We asked for an empty subroute and we return an empty subtoute.  Success!

The use case I'm thinking of is that a user requests a subroute that is [route start, route_end - 50m).  If the route is less than 50m then this would be an empty route, which would not be a failure and would be the intuitive result.

malban:
@elliotjo I think you're right the route points should be cleared in case the output variable wasn't clean to begin with.  Does the rebuildPointIndex function need to get called then?

elliotjo:
Yes, good call on rebuilding the point index.

PR #487, opened 2017-09-28
Add PerpendicularPlaneWithPoint RANSAC model

matt-attack:
This model takes in a point on the plane and a perpendicular axis and tolerance angle from which the plane normal must be to it.

### DIFF
--- a/swri_opencv_util/include/swri_opencv_util/models.h
+++ b/swri_opencv_util/include/swri_opencv_util/models.h
@@ -218,6 +218,38 @@ namespace swri_opencv_util
     float min_angle_;
   };
 
+  class PerpendicularPlaneWithPointFit : public Fit3d<PlaneModel>
+  {
+  public:
+    enum { MIN_SIZE = 2 };
+
+    PerpendicularPlaneWithPointFit(const T& data, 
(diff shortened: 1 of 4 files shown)


swri_opencv_util/src/models.cpp
```diff
@@ -324,6 +324,78 @@ namespace swri_opencv_util
...
+    model.y = p1[1];
+    model.z = p1[2];
+    model.i = normal[0];
+    model.j = normal[1];
+    model.k = normal[2];
+
+    return true;
+  }
+
+  void PerpendicularPlaneWithPointFit::CalculateNorms(const M& model, cv::Mat& norms)
```
malban: Can't this method just be inherited?

matt-attack: It can be and is now, it wasn’t working before but I had some other error which I found after I changed it.

swri_opencv_util/include/swri_opencv_util/models.h
```diff
@@ -218,6 +218,31 @@ namespace swri_opencv_util
     float min_angle_;
   };
 
+  class PerpendicularPlaneWithPointFit : public PlaneFit//Fit3d<PlaneModel>
```
malban: Commented out code

PR #490, opened 2017-10-17
Implement GPU monitor diagnostic node

jatowler:
This is a simple diagnostic node that checks the output of `nvidia-smi` (which obviously assumes an NVidia graphics card) and reports memory and power usage for the GPU(s) on the system to the diagnostics system.

It has no configuration, and it always reports every GPU in the system.


swri_gpu_monitor/package.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<package format="2">
+  <name>swri_gpu_monitor</name>
+  <version>0.0.0</version>
```
pjreed: This version number should be the same as every other package in the repository.

jatowler:
The contents of this package will be moved to a more-appropriate internal package for now, so I'm withdrawing the merge request.

PR #491, opened 2017-10-18
Complete rewrite of initialize_origin.py

### LINKED ISSUE #452: [swri_transform_util] initialize_origin.py Needs to Be Rewritten

evenator:
Add a new Python module `origin_manager` to `swri_transform_util`. `origin_manager` defines a Python class `OriginManager` that can be used to manage the local XY origin. initialize_origin.py is now a wrapper around an OriginManager object.

All tests should pass to maintain previous behavior of initialize_origin.py.

This PR is WIP pending proper documentation, but I'd like to open this to review and CI.

Closes #452

### DIFF
--- a/swri_transform_util/CMakeLists.txt
+++ b/swri_transform_util/CMakeLists.txt
@@ -17,6 +17,8 @@ find_package(catkin REQUIRED COMPONENTS
   topic_tools
 )
 
+catkin_python_setup()
+
 find_package(OpenCV  REQUIRED core highgui imgproc video)
 find_package(Boost REQUIRED)
 
@@ -88,11 +90,8 @@ if(CATKIN_ENABLE_TESTING)
   add_rostest_gtest(test_transform_util launch/transform_util.test test/test_transform_util.cpp)
   target_link_libraries(test_transform_util ${PROJECT_NAME})
 
-  if($ENV{ROS_DISTRO} STREQUAL "indigo")
-    add_rostest(test/initialize_origin_auto_gps.test)
-  else()  # Jade, Kinetic, Lunar, etc
-    add_rostest(test/initialize_origin_auto_navsat.test)
-  endif()
+  add_rostest(test/initialize_origin_auto_gps.test)
+  add_rostest(test/initialize_origin_auto_navsat.test)
   add_rostest(test/initialize_origin_manual.test)
 endif()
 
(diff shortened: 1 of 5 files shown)


evenator:
The test that's failing is setting the origin from NavSatFix because I changed the topic from `gps` to `navsatfix` to allow the node to subscribe to both. Options are:

1. Change the test (and all dependent systems) to use the `navsatfix` topic.
2. Add a parameter to choose which open to subscribe to. Change the test (and all dependent systems) to set the parameter.
3. Add a switch to initialize_origin.py to detect the ROS version and choose which message type to subscribe to on `gps`.

I'm leaning towards `1`, and I'm opposed to `3`.

pjreed:
I'd lean toward #1.  The only reason they were ever different is because there was a long period of time where GPSFix didn't exist in Kinetic, but it does now, so both are perfectly valid.

As a side anecdote, in other nodes  that I've seen use NavSatFix, it seems like `fix` is the commonly used default topic name.  One one hand I like that it's shorter than `navsatfix`, but on the other hand it's not explicitly *not* a GPSFix.

evenator:
I'm ok with `gps` and `fix` for topic names.

evenator:
I'd also like to write a C++ implementation of `OriginManager`, since I've had to write it more than once now while adapting different localization sources.

evenator:
This is now fully documented and passes all tests.

evenator:
Currently, the time stamp of the `PoseStamped` is unset. How should it be set?

pjreed:
I think it'd make sense to set the timestamp based on the header stamp of the received GPSFix or NavSatFix message.

Otherwise, I finally had the chance to read over this, and it looks good to me.

evenator:
Any other opinions on this? @elliotjo or @malban?

facontidavide:
Hi,

thumb up for this change.
I have a question though, that is partially related to Mapviz.

Thanks to the option "auto" and the NavSatFix help with the XY offset between /far_field and my fixed frame, but how do you people at SWRI deal with the orientation offset (yaw) between the coordinate frames?

Thanks in advance

Davide

PR #492, opened 2017-10-21
Normalize TF frames before comparisons.

### LINKED ISSUE #444: swri_transform_util handles "/wgs84" but not "wgs84"
See swri-robotics/mapviz#478.  I would expect `swri_transform_util` to handle whatever its `wgs84_frame` is set to regardless of whether it has a slash in front of it or not.

evenator:
- Add function to normalize leading slash in TF frames
- Add function to compare normalized TF frames
- Use normalized comparisons for all TF frame comparisons

Closes #444

### DIFF
--- a/swri_transform_util/include/swri_transform_util/local_xy_util.h
+++ b/swri_transform_util/include/swri_transform_util/local_xy_util.h
@@ -169,9 +169,9 @@ namespace swri_transform_util
     std::string Frame() const { return frame_; }
 
     /**
-     * Return the TF frame ID corresponding to the local origin with the slash removed
+     * Return the TF frame ID corresponding to the local origin with a leading slash
      *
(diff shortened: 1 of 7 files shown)


evenator:
This is work-in-progress until I can add some tests for the things it fixes.

swri_transform_util/include/swri_transform_util/local_xy_util.h
```diff
@@ -168,6 +169,13 @@ namespace swri_transform_util
     std::string Frame() const { return frame_; }
 
     /**
+     * Return the TF frame ID corresponding to the local origin with the slash removed
```
evenator: Note to self: fix this documentation. This function returns the frame *with* a slash, not without.

evenator:
This should be good to merge.

PR #493, opened 2017-10-27
update to use non deprecated pluginlib macro

mikaelarguedas:
These macros, deprecated for now 8 years, will be removed in the next ROS release (ROS Melodic)

This change will allow the code to keep compiling on future ROS versions

Note: When replacing the macros I noticed you had your own `SWRI_NODELET_EXPORT_CLASS` macro so I modified all nodelets to use it


pjreed:
Thanks again!

PR #496, opened 2017-11-06
Fix pip install of catkin-lint in CI job

evenator:
Add the --user flag so that pip doesn't try to modify root files.


evenator:
Check the output of the Catkin Lint CI job before merging. It previously was failing with a pip error during the setup step. That CI job isn't required, so the PR will pass even with it failing. I expect the CI job will still fail due to actual Catkin lint.

PR #498, opened 2017-11-07
Add new methods that expose the frame timeout.

evenator:
Add two new GetTransform methods to TransformManager that expose the
wait duration for frames to be available for tf::TransformListener frames.

Replace existing GetTransform method with a wrapper around the new method
that calls it with Duration(0.1) to replicate the current behavior and avoid
ABI breakage.

Update doxygen documentation accordingly.


pjreed:
:+1:

PR #499, opened 2017-11-11
keep LocalXyWgs84Util subscribed

facontidavide:
Hi,

I am working on an improvement of mapviz, to be able to change the local_xy_origin directly from the GUI.
I suggest this change, because otherwise there is no way to change an already initialized instance of LocalXyWgs84Util.

Cheers

### DIFF
--- a/swri_transform_util/include/swri_transform_util/local_xy_util.h
+++ b/swri_transform_util/include/swri_transform_util/local_xy_util.h
@@ -214,6 +214,7 @@ namespace swri_transform_util
 
     ros::Subscriber origin_sub_;
     bool initialized_;
+    ros::Time prev_update_;
 
     void Initialize();
 
--- a/swri_transform_util/src/local_xy_util.cpp
+++ b/swri_transform_util/src/local_xy_util.cpp
@@ -130,8 +130,9 @@ namespace swri_transform_util
 
   void LocalXyWgs84Util::HandleOrigin(const topic_tools::ShapeShifter::ConstPtr msg)
   {
-    if (!initialized_)
+    if (!initialized_ || (ros::Time::now() - prev_update_).toSec() > 1.0 )
     {
+      prev_update_ = ros::Time::now();
       ros::NodeHandle node;
       bool ignore_reference_angle = false;
       node.param("/local_xy_ignore_reference_angle", ignore_reference_angle, ignore_reference_angle);
@@ -161,7 +162,6 @@ namespace swri_transform_util
         frame_ = frame;
 
         Initialize();
-        origin_sub_.shutdown();
         return;
       }
       catch (...) {}
@@ -191,7 +191,6 @@ namespace swri_transform_util
         frame_ = frame;
 
         Initialize();
-        origin_sub_.shutdown();
         return;
       }
       catch (...) {}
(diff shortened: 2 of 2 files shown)


evenator:
I'm concerned about possible race conditions and timing weirdness in systems that use this behavior. Imagine two or more nodes whose local origin gets out of sync for even one cycle.

facontidavide:
I see your point. What about adding a ResetInitialization method than that the client code can invoke directly?

facontidavide:
Let me think out loud...

1) The purpose of local_xy_origin is to attach a geolocalization to a fixed frame of the TF.  I understand that you software stack use this information. On the other hand mapviz users like me don't necessary use your entire stack.

2) local_xy_origin is published on a topic. Since it is supposed to be a constant, is,'t __ROS param__ a more suitable tool to set this constant? This would make easier to use launch files to set its value.

3) There are classes such as __LocalXyWgs84Util__ and __TransformManager__ that read this information only once. The local_xy_origin can not be reconfigured unless all the node a shut down. I can hardly decide if this is a bug or a feature ;)

malban:
1.  When mixing geocetric data into ROS/Tf it needs to be converted into a cartesian coordinate system.  The localxy origin/transformation defines one type of projection between geolocated data and a cartesian plane.  Another possible transformation is UTM, which still needs to have the zone and hemisphere defined, which are determined by lat/lon.   In either case it's not clear what the alternative is to defining the origin/parameters that tie the geocentric data to the local Tf coordinate system.

2. Publishing as a topic has the advantage of being easily recordable in bag files so the same state of the system can be reproduced on playback.

3. I think adding a re-initialization method is desirable.  In general I would like to see mapviz be able to continue working correctly across roscore restarts.  I think this would mainly require re-initializing the internal Tf buffer and re-initializing the local origin.

facontidavide:
> I think adding a re-initialization method is desirable. In general I would like to see mapviz be able to continue working correctly across roscore restarts. I think this would mainly require re-initializing the internal Tf buffer and re-initializing the local origin.

This is exactly my purpose, but I should admit that this PR might be a little bit simplistic...
Any suggestion?

malban:
One option would be to add re-initialization functions to the various classes related to the transform manager such as LocalXyWgs84Util, TransformManager, UTMTransformer, and Wgs84Transformer.

A potentially simpler solution might be to just re-instantiate the TransformListener/TransformManger objects in mapviz.  That should have the same effect of causing the local origin to be re-subscribed to and clearing the TF listener cache.

Most of the relevant code is in: https://github.com/swri-robotics/mapviz/blob/indigo-devel/mapviz/include/mapviz/mapviz_plugin.h

It would probably be easiest to change the current tf_manager_ object to be a shared_ptr to make re-instantiating it cleaner.  This would require minor modifications to the plugins which currently  use tf_manager as an object.  An initializeTf function could be added the the plug-in base class to facilitate this.  Many plug-ins also cache transformed versions of the data, so it would probably make sense to make a virtual function to clear data from such plug-ins.

There are also references transform listeners and managers  is in:
https://github.com/swri-robotics/mapviz/blob/indigo-devel/mapviz/include/mapviz/map_canvas.h and  
https://github.com/swri-robotics/mapviz/blob/indigo-devel/mapviz/include/mapviz/mapviz.h

facontidavide:
thanks, I will take a look at that. I guess I will close the PR, for the time being

PR #501, opened 2017-11-14
expose the local_xy_util_

facontidavide:
Hi.

exposing the shared_ptr of local_xy_util_,  plugins in mapviz can use it instead of creating their own instance.

### DIFF
--- a/swri_transform_util/src/transform_manager.cpp
+++ b/swri_transform_util/src/transform_manager.cpp
@@ -363,4 +363,9 @@ namespace swri_transform_util
   {
     return GetTransform(target_frame, source_frame, ros::Time(0), timeout, transform);
   }
+
+  const LocalXyWgs84UtilPtr &TransformManager::LocalXyUtil() const
+  {
+    return local_xy_util_;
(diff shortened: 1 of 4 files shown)


facontidavide:
any feedback about this PR?

swri_transform_util/src/transform_manager.cpp
```diff
@@ -363,4 +363,9 @@ namespace swri_transform_util
   {
     return GetTransform(target_frame, source_frame, ros::Time(0), timeout, transform);
   }
+
+  LocalXyWgs84UtilPtr &TransformManager::local_xy_util()
```
pjreed: This looks good to me, although to be consistent with the rest of the class, it would be better for this to be named using upper camel case, so `TransformManager::LocalXyUtil()`.

facontidavide:
agree about the name. I am also considering to make it constant

facontidavide:
sorry, I messed up with the rebase. do not merge it

facontidavide:
Fixed

PR #503, opened 2017-11-27
Add swri_dbw_interface package.

malban:
Add swri_dbw_interface package.


kriskozak:
I think that the documentation was fine as-is for internal use, but it needs a little bit of modification for public consumption. I'll comment in-line.

swri_dbw_interface/doc/package.md
```diff
@@ -0,0 +1,366 @@
...
+   tha command topic, or not subscribe to the topic at all.
+
+1. Use the ROS console macros to provide useful feedback for
+   developers and troubleshooters, but be careful not spam the output.
+
+
+## Namespace [⬆](#swri_dbw_interface)
+
+By convention, all topics related to the drive-by-wire system are in
+the "/vehicle_interface" namespace.
```
kriskozak: Use of this namespace is one reason why I preferred the name swri_vehicle_interface to swri_dbw_interface.  Is there any value to using /dbw_interface as the namespace instead?  In any case, I think it would be beneficial to segregate the swri_dbw_interface topics from any native dbw ROS topics by namespace... perhaps /dbw_interface would be a good place to put those.

swri_dbw_interface/doc/package.md
```diff
@@ -0,0 +1,366 @@
...
+
+
+## Throttle Interface [⬆](#swri_dbw_interface)
+
+The throttle interface has two main topics for control:
+
+- **/vehicle_interface/throttle_command** -
+*(marti_common_msgs/Float64Stamped)* - *subscription* - Desired
+position of the throttle axis normalized between [0.0, 1.0].  Expected
+at a constant rate from the high level ROS system (typically 50Hz).
```
kriskozak: I still think it's a good idea to not be wishy washy on the rates here.  Specifying a rate (and perhaps tolerances) facilitates routine monitoring of health. 

The same goes for timeouts. 

Any vehicles for which the standard values don't work, can be customized, and treated as an exception to the standard.

kriskozak: There's not an obvious place to put this comment, but it vaguely relates to my previous comment about the rate and timeout monitoring...

I think there should be some discussion on error handling and/or failure modes. For example, what should happen if a steering command times out?  This is not a simple modification, so it's probably fine to merge this before addressing this issue, but I think it's important to include.

swri_dbw_interface/include/swri_dbw_interface/device_states.h
```diff
@@ -0,0 +1,64 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+// The follow define common states used for transmission devices in the
+// swri_dbw_interface.
+
+// These values are valid command and feedback states.
+static const std::string TRANSMISSION_PARK = "park";
+static const std::string TRANSMISSION_DRIVE = "drive";
+static const std::string TRANSMISSION_REVERSE = "reverse";
+static const std::string TRANSMISSION_NEUTRAL = "neutral";
```
kriskozak: Does it make any sense to incorporate more possible states here, given that some others, particularly drive_low, may be available on most vehicles?

swri_dbw_interface/doc/package.md
malban: I prefer /vehicle/ or /vehicle_interface/ to /dbw_interface/

swri_dbw_interface/include/swri_dbw_interface/device_states.h
malban: I added drive_low

swri_dbw_interface/doc/package.md
```diff
@@ -0,0 +1,366 @@
...
+are encountered.
+
+Due to the combined unique needs of projects and variations in
+drive-by-wire capabilities and implementations, it is expected that
+variations and adaptations may be required for different
+circumastances.  In those cases, any variations should be documented
+along with the implementation.
+
+As new features are required, it is recommended that they be developed
+for the specific project according to the conventions below, tested
```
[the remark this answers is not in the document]
malban: Replaced with "use case"

swri_dbw_interface/doc/package.md
```diff
@@ -0,0 +1,366 @@
...
+The important aspect is a consistent interface presented to the higher
+level autonomy system, not how it is implemented.
+
+This interface has evolved against actuated gas/brake
+Ackermann-steered vehicles (as opposed to tracked vehicles or vehicles
+that provide intrinsic speed control).  It is expected that the
+interface will be extended to handle other types of vehicles as they
+are encountered.
+
+Due to the combined unique needs of projects and variations in
```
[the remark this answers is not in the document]
malban: Replaced with "use cases"

swri_dbw_interface/doc/package.md
kriskozak: I agree, and I prefer /vehicle_interface to /vehicle. Does it make sense to you to segregate native DBW messages in a different namespace? I know we mix them on some (all?) platforms, and I find it difficult to distinguish which are which at a glance. Also, there's the (small) possibility that naming conventions may conflict.

swri_dbw_interface/doc/package.md
```diff
@@ -0,0 +1,366 @@
...
+   (or the predicted state if no actual feedback is available).
+
+1. Command topics should have topic names with the postfix "_command".
+
+1. Feedback topics should have topic names with the postfix "_feedback".
+
+1. Other topics should be given names that make it obvious whether
+   they are an input to or output from the DBW system.
+
+1. Prefer to use strings (with constants) over integer enumerations so
```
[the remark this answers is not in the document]
kriskozak: This is fine, but I would suggest adding a more specific callout to where to look for the constants (swri_dbw_interface/device_states.h:). Adding it as its own item in this list would be fine.

kriskozak:
Aside from the one new recommendation to point out the file with the constants at the top (I know that it shows up in the relevant sections below), it looks good to me.

PR #504, opened 2017-12-04
Use pkgconfig to include libproj in swri_transform_util

pjreed:
Previously the CMakeLists.txt simply appended "proj" to the library
list, which happened to work, but is not actually the correct way to
include libraries that provide a .pc file; that could potentially fail
on any systems that don't have libproj in their default include path.

Discovered by @matt-attack.

### DIFF
--- a/swri_transform_util/CMakeLists.txt
+++ b/swri_transform_util/CMakeLists.txt
@@ -20,15 +20,14 @@ find_package(catkin REQUIRED COMPONENTS
 find_package(OpenCV  REQUIRED core highgui imgproc video)
 find_package(Boost REQUIRED)
 
-if($ENV{ROS_DISTRO} STRLESS "kinetic")
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(PROJ proj)
+if(NOT "${PROJ_FOUND}")
   set(PROJ_INCLUDE_DIRS /usr/include)
   set(PROJ_LIBRARY_DIRS /usr/lib)
   find_library(PROJ_LIBRARIES proj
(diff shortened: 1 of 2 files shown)


pjreed:
Well this is sad.  `libproj-dev` apparently only includes a pkgconfig file on Indigo, not Kinetic...

pjreed:
On Indigo it will instead use `find_library`, which is less ideal than pkg-config but still preferable to just adding a raw library name.

swri_transform_util/CMakeLists.txt
```diff
@@ -20,6 +20,17 @@ find_package(catkin REQUIRED COMPONENTS
 find_package(OpenCV  REQUIRED core highgui imgproc video)
 find_package(Boost REQUIRED)
 
+if($ENV{ROS_DISTRO} STRLESS "kinetic")
```
evenator: I think it would be better here to switch on the actual failure (pkg-config), rather than the ROS version. It would be more robust and less dependent on the ROS version correlating wtih the OS. Something like:

```cmake
find_package(PkgConfig REQUIRED)
pkg_check_modules(PROJ QUIET proj)
if (NOT ${PROJ_FOUND})
  set(PROJ_INCLUDE_DIRS /usr/include)
  set(PROJ_LIBRARY_DIRS /usr/lib)
  find_library(PROJ_LIBRARIES proj
    HINTS ${PROJ_LIBRARY_DIRS}
  )
endif()
```

pjreed: Good point.  Changed.

PR #505, opened 2017-12-13
Ensure all swri::Subscriber members are initialized

pjreed:
Some member variables in swri::Subscriber were uninitialized by default; this ensures all of them are initialized.

### DIFF
--- a/swri_roscpp/include/swri_roscpp/subscriber_impl.h
+++ b/swri_roscpp/include/swri_roscpp/subscriber_impl.h
@@ -75,8 +75,7 @@ class SubscriberImpl
       min_latency_ = latency;
       max_latency_ = latency;
       total_latency_ = latency;
-    }
-    else {
+    } else {
       min_latency_ = std::min(min_latency_, latency);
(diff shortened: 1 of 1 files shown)


pjreed:
As long as I'm in here, looking at subscriber_impl.h:73, it appears that if the stamp in a header is 0, this will very quickly cause an overflow of `total_latency_`.  Maybe `processHeader` should skip the latency check if  the header's stamp isn't set?

swri_roscpp/include/swri_roscpp/subscriber_impl.h
```diff
@@ -75,7 +75,8 @@ class SubscriberImpl
       min_latency_ = latency;
       max_latency_ = latency;
       total_latency_ = latency;
-    } else {
+    }
+    else {
```
elliotjo: Recommend undoing these style changes to remain consistent.

pjreed: Any opinions on how that check should be handled if the header stamp isn't set?

evenator:
LGTM

PR #513, opened 2018-05-09
Catch and fix NaNs in the fillOrientations function

kriskozak:
- Check for NaNs in the swri_route_util/src/util.cpp fillOrientations function
- Replace the invalid orientations with orientations from nearest neighbors
  in the route with valid orientations
- Simplify the orientation calculation to use center difference rather than
  averaging of vectors


pjreed:
I'm going to make a new release of marti_common soon, should we get this change in first?

PR #519, opened 2018-07-11
Handle invalid fixes properly

### LINKED ISSUE #518: swri_transform_util/initialize_origin.py unsubscribes from GPSFix/NavSatFix topics even if they're invalid
If the initialize_origin.py script gets a GPSFix or NavSatFix mesage that is invalid, it logs the fact that it's invalid, but then it unsubscribes from the topic anyway and can never get a valid one.

pjreed:
If initialize_origin.py got an invalid fix, it would unsubscribe from
the fix/gps topics but not publish a /local_xy_origin, thus leaving the
system eternally stranded.  Now it will continue to listen until a valid
fix arrives.

Fixes #518


matt-attack:
Looks pretty good! +10 for test cases

PR #520, opened 2018-07-23
Update package maintainers

pjreed:


swri_rospy/package.xml
```diff
@@ -6,7 +6,7 @@
   This package provides added functionaliy on top of rospy, including a
   single-threaded callback queue.
   </description>
-  <maintainer email="evenator@swri.edu">Ed Venator</maintainer>
```
evenator: 😞

pjreed: I'm assuming you don't have access `evenator@swri.edu` any more, but feel free to add a different e-mail address if you still want to claim responsibility for it. :wink:

PR #522, opened 2018-09-10
Add function to blend two images with alpha channels.

malban:


danthony06:
Might be nice if the code in swri_image_util used this functionality.

https://github.com/swri-robotics/marti_common/blob/master/swri_image_util/src/blend_images_util.cpp

malban:
@danthony06 They seem to do slightly different things.  Though I suppose you could create two alpha channels based on the logic in blend_images_util and then call this function.  

The base image would have an alpha channel with (1.0 - alpha) for all pixels.  The top image would have an alpha channel of alpha for all pixels in the mask and 0.0 elsewhere.

Right?

danthony06:
Yes, I think that would do it.

PR #525, opened 2018-10-22
Add function to test if two line segments intersect.

malban:


swri_geometry_util/include/swri_geometry_util/intersection.h
```diff
@@ -51,6 +51,41 @@ namespace swri_geometry_util
       cv::Vec2d p4,
       cv::Vec2d& c);
 
+  /**
+   * Calculate the instersection between two line segments defined by 4 points.
+   *
+   * In the case of parallel overlapping segments, the intersection point
+   * closest to p1 is returned.
+   *
+   * @param[in]   p1   First point of line segemnt 1.
```
pjreed: Typo in `segemnt`.

swri_geometry_util/include/swri_geometry_util/intersection.h
```diff
@@ -51,6 +51,41 @@ namespace swri_geometry_util
...
+   * @param[in]   p2   Second point of line segemnt 1.
+   * @param[in]   p3   First point of line segemnt 2.
+   * @param[in]   p4   Second point of line segemnt 2.
+   * @param[out]  c    The intersection point.
+   *
+   * @returns True if the line segments intersect.
+   */
+  bool LineSegmentIntersection(
+      cv::Vec2d p1,
```
pjreed: All of the in-arguments to both this and `PointOnLineSegment` can probably be const references.

swri_geometry_util/include/swri_geometry_util/intersection.h
malban: Fixed typo in this function and the existing function above it.

swri_geometry_util/include/swri_geometry_util/intersection.h
malban: Fixed the parameters in this function and the existing function above it.

PR #527, opened 2018-10-31
Transform the polygon points of an obstacle based off the pose of the…

agyoungs:
… obstacle

### DIFF
--- a/swri_route_util/src/route_speeds.cpp
+++ b/swri_route_util/src/route_speeds.cpp
@@ -316,7 +316,6 @@ void generateObstacleData(
     for (size_t j = 0; j < obs_msg.polygon.size(); j++) {
       tf::Vector3 pt;
       tf::pointMsgToTF(obs_msg.polygon[j], pt);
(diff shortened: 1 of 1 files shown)


swri_route_util/src/route_speeds.cpp
```diff
@@ -316,7 +316,6 @@ void generateObstacleData(
     for (size_t j = 0; j < obs_msg.polygon.size(); j++) {
       tf::Vector3 pt;
       tf::pointMsgToTF(obs_msg.polygon[j], pt);
-      pt = pt - g_obs_from_local.getOrigin();
```
malban: seem to be missing the actual transform part

malban: never mind, its already there, you're just removing the incorrect translation

agyoungs: Yea, I just removed the bad translation.  The translation would have no effect when an identity pose was used, but this didn't work if the polygon was relative to the pose.  I tested this change against current usages and it did not change the behavior.

PR #532, opened 2018-12-10
Add dynamic parameters

matt-attack:
This adds a class that handles dynamic configuration of parameters in a much cleaner way than with standard dynamic_reconfigure. 

It also allows for setting the list of configurable parameters are runtime rather than compile time.


matt-attack:
@pjreed looks like we had another test failure. Looking at the test I'm not really sure what caused it as there doesn't appear to be a way for that publisher to be closed.

pjreed:
How does this work?  Could you add some documentation for it?

Seems like the test failure must be timing-based.  Re-running it eventually makes it pass.  Sometime I'll investigate and see if I can figure out what's causing it...

matt-attack:
Yeah, can do. 

It implements the same interface as the dynamic_reconfigure system, so it works with all the same tools, but sets up the configuration at run time in the code instead of at build time in the cfg files. 

Right now, it handles changes by directly changing variables in your class, so you need to make sure they persist and to lock the mutex if you don't want them changing out from under you. It's a much simpler interface, but you need to be careful. Adding a callback for a change instead would also be easy to do for cases where that would be useful.

matt-attack:
@pjreed added a description if you want to take a look. I changed the swri_transform_util dynamic publisher to use this as an example if you would like to see how it works.

matt-attack:
@pjreed Does this look good to merge?

swri_roscpp/include/swri_roscpp/dynamic_parameters.h
```diff
@@ -0,0 +1,450 @@
+#ifndef SWRI_ROSCPP_DYNAMIC_PARAMETERS_H_
```
pjreed: This needs a license header.

swri_roscpp/include/swri_roscpp/dynamic_parameters.h
```diff
@@ -0,0 +1,450 @@
...
+    ros::NodeHandle nh_;
+
+    std::map<std::string, DynamicValue> values_;
+
+    boost::mutex mutex_;
+
+    bool setConfigCallback(dynamic_reconfigure::Reconfigure::Request &req,
+          dynamic_reconfigure::Reconfigure::Response &rsp)
+    {
+      ROS_INFO("Got configuration change request");
```
pjreed: I'm not sure whether I'd rather have this log level be `INFO` or `TRACE`.  Would everybody who has reconfigurable parameters want this to be printed every time something is changed?

swri_roscpp/include/swri_roscpp/dynamic_parameters.h
```diff
@@ -0,0 +1,450 @@
...
+
+        if (iter->second.type != DynamicValue::Double && iter->second.type != DynamicValue::Float)
+        {
+          ROS_ERROR("Value '%s' was not a double type.", param.name.c_str());
+          continue;
+        }
+
+        if (iter->second.type == DynamicValue::Double)
+        {
+          double* v = (double*)iter->second.pointer;
```
pjreed: Any reason to use C-style casts here rather than `static_cast`s?

swri_roscpp/readme.md
```diff
@@ -31,3 +31,51 @@ add_topic_service_files(DIRECTORY topic_srv FILES
...
+{
+  // your node stuff here
+
+  ros::NodeHandle pnh("~");
+  swri::DynamicParameters params;
+  params.initialize(pnh);
+...
+```
+
+You can then start declaring and reading in configuration values, giving a long living pointer to a variable for each. This variable is changed on a dynamic reconfigure of that parameter. If you want to ensure no changes are applied during an operation, be sure to lock the mutex you can get by calling the `params.mutex()` function.
```
pjreed: I think I'd prefer this to say that you should *always* lock the mutex around getting the parameters, and also include and example of that below.  There are some cases where you can get away without locking it, but I'd rather be safe than sorry and leave figuring that out to people who really know what they're doing.

swri_transform_util/src/nodelets/dynamic_publisher.cpp
```diff
@@ -68,43 +72,29 @@ class DynamicPublisher : public nodelet::Nodelet
...
-    {
-      config_ = config;
-    }
-  }
-
   void Publish(const ros::TimerEvent& unused)
   {
     tf::Transform transform;
-    tf::Vector3 origin(config_.x, config_.y, config_.z);
+    tf::Vector3 origin(x_, y_, z_);
```
pjreed: The mutex should be locked around reading these values, since you have six values here that could all be reconfigured, and you wouldn't want any of them to change while you're in the middle of reading them; otherwise you'll end up publishing a transform that never really existed.

pjreed:
I think it'd also be useful to be able to register a callback in order to be notified whenever any values are reconfigured.  I can think of two scenarios where I'd want that:
1. If there's some expensive operation that you only want to perform when a value actually changes
2. If you want to cache the reconfigurable values in non-volatile variables so that you can access them repeatedly without needing to lock a mutex

swri_roscpp/readme.md
```diff
@@ -31,3 +31,51 @@ add_topic_service_files(DIRECTORY topic_srv FILES
...
+
+```cpp
+  float flt;
```
pjreed: Setting the params like this is actually very dangerous.  The DynamicParameters class is storing a reference to this variable, but this variable will be destroyed when the function calling it goes out of scope.  The next time your parameters are reconfigured after that, it will cause a segfault.

I think it'd be better to make wrapper types (`BoolParam`, `IntParam`, etc.) that are actually `shared_ptr`s under the hood in order to make sure they always exist; it would also be possible to instrument those types with getters that could handle automatically locking them to make sure they're thread-safe.

swri_roscpp/include/swri_roscpp/dynamic_parameters.h
pjreed: Scratch that, after taking a break for a few weeks I forgot that ROS didn't have a `TRACE` level.  I guess `DEBUG` works, too.

swri_roscpp/include/swri_roscpp/dynamic_parameters.h
pjreed: For all of the cases here, I'd prefer to use `static_cast` so that it's explicit what type of casting you're doing and also because the compiler is better at catching errors if you're using it wrong.  (see http://www.stroustrup.com/bs_faq2.html#static-cast for more details)

swri_roscpp/readme.md
matt-attack: @pjreed went and added something like this

matt-attack:
If you think it is okay, I might leave adding the callbacks for another issue.

matt-attack:
Hmm. I seem to have accidentally merged in the wrong master causing these issues

matt-attack:
Should be fixed now

matt-attack:
@daniel-dsouza added a function to let the user do that

daniel-dsouza:
looks good to me

daniel-dsouza:
@pjreed have you resolved your review?

PR #534, opened 2018-12-18
Fix the unit test that intermittently breaks

pjreed:


pjreed:
Previously it seemed like the test in question had about a 50% chance, and it's now built & run without failures a total of 16 times, so I'm going to hope this has fixed it.

PR #540, opened 2019-04-26
Add callback for on change for dynamic parameters

matt-attack:
This lets them be used in a similar fashion to standard dynamic reconfigure.


daniel-dsouza:
I approve. @pjreed can you hit the button?

PR #541, opened 2019-04-30
Improve topic service client

nick-alton:
This improves the performance for the topic service client. Specifically fixes a problem that sequences get confused when doing multiple requests from a single node.


swri_roscpp/include/swri_roscpp/topic_service_client.h
```diff
@@ -55,7 +55,7 @@ class TopicServiceClientRaw
   int sequence_;
 
 public:
-  TopicServiceClientRaw() : sequence_(0), timeout_(ros::Duration(4.0))
+  TopicServiceClientRaw() : sequence_(rand()), timeout_(ros::Duration(4.0))
```
pjreed: What's the purpose behind putting a random number in here?

matt-attack: It ensures that different clients making the same service request in the same node start with requests with different sequence numbers so there's less possible confusion between them.

Perhaps this random number would do better to be appended to the client name in the request instead?

swri_roscpp/include/swri_roscpp/topic_service_client.h
```diff
@@ -64,7 +63,12 @@ class TopicServiceClientRaw
                 const std::string &service,
                 const std::string &client_name = "")
   {
-    name_ = client_name.length() ? client_name : ros::this_node::getName();
+    int random_int = rand()%10000;
```
pjreed: I'd rather not use a random number from 0 to 9999 for this since there's a possibility of a collision (especially if you're not seeding the RNG).  If the purpose of this is just to have a unique name, adding a UUID would be more reliable; if it's useful for this to be a relatively small number, I'd rather have something like a static variable that gets incremented every time it's used.

matt-attack: We tried to do that, but ran into the issue where we couldn't figure out where to put a static variable

matt-attack: We might have to do the UUID approach.

evenator: You totally nerd-sniped me on this one.

```c++
struct MonotonicCounter
{
  static uint64_t next()
  {
    ++value_;
    return value_;
  }

  private:
  static uint64_t value_ = 0;
}
```

Usage
```c++
if (client_name.empty()
{
  std::ostringstream ss;
  ss << ros::this_node::getName()  << MonotonicCounter::next();
  name_ = ss.str()
}
else
{
  name_ = client_name;
}
```

matt-attack: :(
`swri_roscpp/include/swri_roscpp/topic_service_client.h:49:28: error: ISO C++ forbids in-class initialization of non-const static member ‘swri::MonotonicCounter::value_’
   static uint64_t value_ = 0;`

evenator: C++03 makes static initialization hard.

PR #542, opened 2019-05-22
 Add an .editorconfig file

KindDragon:
Could help to maintain consistent coding styles

What is .editorconfig:
http://editorconfig.org/#overview

Plugins for IDEs to support:
http://editorconfig.org/#download


pjreed:
I hadn't heard of this before, but this looks pretty neat.  Thanks!

PR #544, opened 2019-05-22
Support in initialize_origin.py specifying custom ROS topic in parameter

KindDragon:
User can specify in parameter 'local_xy_custom_topic' ROS topic that will be used for origin.
Message can use fields with names 'latitude', 'longitude', 'altitude' or 'lat', 'lon' and 'height'.
These fields can also be inside the message field with name 'position'.

Also, you can specify topic for GPSFix message in parameter '~local_xy_gpsfix_topic', and
topic for NavSatFix message in parameter '~local_xy_navsatfix_topic'


pjreed:
Sorry for taking so long to get to this, since it's a big diff I wanted to take the time to read through it all and it just slipped my mind.  This looks really useful, though, thanks!

PR #546, opened 2019-06-21
Fix get relative transform

matt-attack:
The GetRelativeTransform function appears to have not been working. This change makes it behave as expected.


pjreed:
Would it be possible to add a unit test to `test/test_transform_util.cpp` to verify that it's working right now?

matt-attack:
I can pull a case from what I tested it with I suppose

matt-attack:
Added!

PR #548, opened 2019-07-22
Prevent swri:Subscriber latency calculations if current time is zero

jgassaway:
This adds a check to prevent latency calculations if the current ros::Time::now() is zero, which can result in negative latency values that cause ros::Duration to throw an exception and cause the node to crash.


jgassaway:
I have tested this on a bagfile successfully, but need to test it on live hardware.

pjreed:
I think this is working fine, so I'm going to go ahead and merge it.

Issue #562, opened 2019-12-31
Topic Service Generation Gets Run Every Build


matt-attack:
Due to how CMake works the custom target on https://github.com/swri-robotics/marti_common/blob/master/swri_roscpp/cmake/swri_roscpp-extras.cmake.em#L47 that generates the message gets run every single build. This can cause a chain of extra rebuilding throughout a codebase.

Propose changing to use a pattern similar to https://github.com/matt-attack/pubsub/blob/master/tools/CMakeLists.txt#L39

matt-attack:
After poking around at this for a while, I didn't have any luck avoiding this. 

It seems actionlib gets around this by running the generation at build time entirely. They just use safe_execute_process which might avoid some complexity.

matt-attack:
@pjreed do you think moving to that would be an acceptable solution? ROS essentially forces this with message dependency changes anyways so it doesn't add much of a requirement.

In testing with it, it seems to vastly improve rebuild times. (From 16 seconds to 8 seconds for just the message package)

PR #563, opened 2019-12-31
Refactor Topic Service Generation

matt-attack:
Use safe_execute_process to generate messages instead and prevent it from running every build to vastly improve build times.

Resolves https://github.com/swri-robotics/marti_common/issues/562


matt-attack:
Not sure why it says the build failed in Bionic. Didn't see any errors.

matt-attack:
@pjreed can you take another look at this? This would be useful to have merged for build times.

PR #564, opened 2020-02-03
Add Enum Support

matt-attack:
Also allows preserving declaration order of parameters instead of having them displayed in alphabetical order.


swri_roscpp/include/swri_roscpp/dynamic_parameters.h
```diff
@@ -392,6 +426,34 @@ namespace swri
         desc.level = 0;
         desc.description = param->second.description;
         desc.edit_method = "";
+
+        // If this is an enum, lets make the edit method string
+        for (int i = 0; i < param->second.enums.size(); i++)
```
pjreed: This looks like it is shadowing the `i` parameter from the outer loop, is that intentional?  And is there a reason to not use for-each loops here?

swri_roscpp/include/swri_roscpp/dynamic_parameters.h
```diff
@@ -59,6 +59,7 @@ namespace swri
     Type type;
     std::string name;
     std::string description;
+    std::vector<std::pair<std::string, int>> enums;
```
pjreed: Not having a space between the `>`'s will break some versions of gcc, although I'm not sure if that's an issue on Ubuntu 16.04 / 18.04.
```suggestion
    std::vector<std::pair<std::string, int> > enums;
```

swri_roscpp/include/swri_roscpp/dynamic_parameters.h
```diff
@@ -392,6 +426,34 @@ namespace swri
         desc.level = 0;
         desc.description = param->second.description;
         desc.edit_method = "";
+
+        // If this is an enum, lets make the edit method string
+        for (int i = 0; i < param->second.enums.size(); i++)
+        {
+          if (i == 0)
+          {
+            desc.edit_method += "{'enum_description': '" + escapeString(desc.description) + "'";
```
pjreed: Assembling YAML strings manually seems dangerous, especially writing our own method to escape strings.  Why not use yaml-cpp?

swri_roscpp/include/swri_roscpp/dynamic_parameters.h
```diff
@@ -405,6 +467,20 @@ namespace swri
 
       set_service_ = nh_.advertiseService("set_parameters",
             &DynamicParameters::setConfigCallback, this);
+
+      ordered_params_.clear();// to save memory
+    }
+
+    void addEnums(const std::string& param, const std::vector<std::pair<std::string, int>>& enums)
```
pjreed:
```suggestion
    void addEnums(const std::string& param, const std::vector<std::pair<std::string, int> >& enums)
```

swri_roscpp/include/swri_roscpp/dynamic_parameters.h
matt-attack: Ooops. Should not do that.

I wanted an index so writing out the loop manually is usually the most concise way of doing that.

matt-attack: Also, it seems some people are requesting this to support C++03 which doesnt allow those.

swri_roscpp/include/swri_roscpp/dynamic_parameters.h
matt-attack: Was going for the most prudent solution and didn't want to add extra dependencies.

HoomanBean:
When is this gonna get merged. I need it.

PR #576, opened 2020-04-16
swri_geometry_util: fix build with GEOS 3.8

lopsided98:
`swri_geometry_util` fails to build with GEOS 3.8 due to the following commit: https://github.com/libgeos/geos/commit/7ee79a3f3d866623176ed1d4c5d164a58377f825.

I think this fix is compatible with older versions of GEOS, but I haven't tested.


pjreed:
This looks good, thanks!

PR #583, opened 2020-06-12
Use setuptools instead of distutils

sloretz:
Use setuptools instead of distutils ros/catkin#1048

In Ubuntu Focal `distutils` is not installed by default, so it might not be available. `ros-noetic-catkin` depends on `python3-setuptools`, so we know setuptools will be available.

This PR assumes `master` targets only ROS Noetic. If it actually covers kinetic and melodic, then both `setuptools` should be used and the `package.xml` for this package should be edited as described in the migration guide:

http://wiki.ros.org/noetic/Migration#Setuptools_instead_of_Distutils

```xml
  <buildtool_depend condition="$ROS_PYTHON_VERSION == 2">python-setuptools</buildtool_depend>
  <buildtool_depend condition="$ROS_PYTHON_VERSION == 3">python3-setuptools</buildtool_depend>
```

### DIFF
--- a/marti_data_structures/CMakeLists.txt
+++ b/marti_data_structures/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.0.2)
 project(marti_data_structures)
 
 find_package(catkin REQUIRED)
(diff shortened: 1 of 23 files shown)


pjreed:
The master branch does support all three releases, so I'll go ahead and make that change.  Thanks!

PR #594, opened 2020-06-30
Improve topic service macro on ROS2

matt-attack:
Previously it would not work if you also generated other messages in the same project.


matt-attack:
@pjreed this has been tested and should be good to go. Not sure why only those builds are failing or how to re-run them.

pjreed:
I think those builds only ran in the first place because this PR originally targeted the `master` branch; those are doing pull request testing buildings for Kinetic/Melodic/Noetic, which are all ROS1 so they don't matter for this PR anyway.

PR #595, opened 2020-07-14
Added python support for wgs84 transforms

agyoungs:
This gives basic functionality to convert local coordinates to WGS84 coordinates and vice versa in python.

PR #596, opened 2020-07-14
Add python wgs84 transformer ros2

agyoungs:
Python WGS84 Transformer from #595

### DIFF
--- a/swri_transform_util/src/swri_transform_util/wgs84_transformer.py
+++ b/swri_transform_util/src/swri_transform_util/wgs84_transformer.py
@@ -99,7 +99,7 @@ class Wgs84Transformer(object):
 
         d = points.dot([[self._sin_heading * self._cos_heading, self._cos_heading],
                         [self._sin_heading**2, self._sin_heading]])
-        d[:,0] = (points[:,1] - d[:,0]) / self._cos_heading
+        d[:, 0] = (points[:, 1] - d[:, 0]) / self._cos_heading
 
         r = d / [self._rho_lat, self._rho_lon] + [self._reference_latitude, self._reference_longitude]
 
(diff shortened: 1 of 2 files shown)


swri_transform_util/src/swri_transform_util/wgs84_transformer.py
```diff
@@ -0,0 +1,112 @@
...
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import math
+import numpy as np
+from tf.transformations import euler_from_quaternion
```
pjreed, line 35: `tf` doesn't exist in ROS 2, so you'll need to find out if either `tf2` has an equivalent to this or go find the original function and copy-paste it in here.

agyoungs: I couldn't find a standard ROS 2 implementation, so I added a simple yaw_from_quaternion method. I also put in a TODO to update to a more standard implementation when available.

pjreed:
Cool, thanks!

PR #600, opened 2020-09-08
stamp local_xy_origin message picked from list with now()

daniel-dsouza:


pjreed:
LGTM; do you know if this is also an issue for the ROS2 version in the `dashing-devel` branch?  If so, could you cherry-pick this over there?

daniel-dsouza:
> LGTM; do you know if this is also an issue for the ROS2 version in the `dashing-devel` branch? If so, could you cherry-pick this over there?

no idea. Currently don't have an environment for ROS2 setup.

matt-attack:
> > LGTM; do you know if this is also an issue for the ROS2 version in the `dashing-devel` branch? If so, could you cherry-pick this over there?
> 
> no idea. Currently don't have an environment for ROS2 setup.

You should be able to just cherry-pick it onto it without trying to test or build it.

daniel-dsouza:
> > > LGTM; do you know if this is also an issue for the ROS2 version in the `dashing-devel` branch? If so, could you cherry-pick this over there?
> > 
> > 
> > no idea. Currently don't have an environment for ROS2 setup.
> 
> You should be able to just cherry-pick it onto it without trying to test or build it.

Don't think we have an `origin_manager.py` for ros2

PR #602, opened 2020-09-14
issue-601-topic-services-under-ROS2

dvmartin999:
Modify service splitter to use snake_case names as required by ROS2; remove unused include files


swri_roscpp/include/swri_roscpp/swri_roscpp.h
```diff
@@ -31,8 +31,8 @@
 
 #include <swri_roscpp/publisher.h>
 #include <swri_roscpp/subscriber.h>
-#include <swri_roscpp/latched_subscriber.h>
-#include <swri_roscpp/parameters.h>
+//#include <swri_roscpp/latched_subscriber.h>
+//#include <swri_roscpp/parameters.h>
```
pjreed, line 35:
```suggestion
```
Might as well just remove them, no need to leave them in but commented out.

swri_roscpp/scripts/service_splitter.py
```diff
@@ -43,6 +43,11 @@
 data = sys.argv
 
 
+# https://stackoverflow.com/questions/1175208/elegant-python-function-to-convert-camelcase-to-snake-case
```
danthony06, line 46: My only concern with this is whether or not we strictly obey camel case. Is it possible that we end up feeding something like "GPS" into this, and get g_p_s out? Is it worth it to convert to camel case, or should we just leave it as is?

dvmartin999: I don't think leaving it as-is is an option.  As noted in the bug report, generated topic services simply don't compile, since the underlying message generation in ROS2 switches file names to camelcase.  It's not just a style issue. That said, I agree that a different function to apply camel case conversion might make sense, particularly if we re-use whatever the ROS2 message generation is using, then it's more likely to be consistent.

danthony06: I changed this in my PR here: https://github.com/swri-robotics/marti_common/pull/604 to use the same logic as the rosidl package does when converting from camel to snake case.

danthony06:
This was fixed as part of: https://github.com/swri-robotics/marti_common/pull/604

PR #603, opened 2020-09-21
Fix header warnings

danthony06:
I fixed a bunch of includes to remove deprecated headers to reduce the number of build warnings.

Distro A, OPSEC #4584

### DIFF
--- a/swri_image_util/src/nodes/blend_images_node.cpp
+++ b/swri_image_util/src/nodes/blend_images_node.cpp
@@ -31,7 +31,7 @@
 
 #include <rclcpp/rclcpp.hpp>
 #include <cv_bridge/cv_bridge.h>
-#include <image_transport/image_transport.h>
+#include <image_transport/image_transport.hpp>
 #include <message_filters/subscriber.h>
 #include <message_filters/time_synchronizer.h>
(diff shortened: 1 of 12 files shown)


pjreed:
While these deprecation warnings are very annoying, unfortunately, the new versions of these headers did not exist prior to Foxy.  Until we're ready to drop support for Dashing and Eloquent, I'd rather just deal with warning messages than unnecessarily break compatibility.  Is there any other way to suppress those warnings?

danthony06:
Hmm, I can check into that. I don't suppose we want a separate Foxy branch, do we?

pjreed:
I'd rather not have a separate branch if we could avoid it, since that means every PR would need to be cherry-picked into an additional branch.  It'd actually be nice if we only had a single `master` branch that worked in both ROS1 and ROS2, which is possible, but that would require a ton of conditional compilation stuff that we didn't want to deal with...

danthony06:
Okay, I'll look into other methods of suppressing the warning or another
way of solving it.

danthony06:
Closing this in favor of: https://github.com/swri-robotics/marti_common/pull/621

PR #604, opened 2020-09-22
Issue 601 topic services under ros2

danthony06:
Distro A, OPSEC #4584


pjreed:
Just checking in since it's been a while since I saw any updates for this, is it ready to go?

danthony06:
I think so, but I don't have a great test for it.

dvmartin999:
@danthony06 I was able to pull your branch and confirm it worked in my test case that generated the error in the first place.

Without the fix, my code fails to compile with errors like: "<package>/srv/DeleteRoute.h:4:10: fatal error: <package>/msg/DeleteRouteResponse.hpp: No such file or directory
    4 | #include <package>_msgs/msg/DeleteRouteResponse.hpp

With the fix, my code compiles and runs correctly.

@pjreed Is that sufficient validation to move this patch forward?  Thank

pjreed:
You bet, that looks good to me.  Thanks!

dvmartin999:
Good deal @pjreed !  Any guidance on when we can expect this fix to be rolled into the distribution that's accessed via package managers on Ubuntu? Just need to give the CI guys a heads-up on what to expect...

pjreed:
I've tagged a new release and opened PRs to get it released into Dashing, Eloquent, and Foxy:
https://github.com/ros/rosdistro/pull/27468
https://github.com/ros/rosdistro/pull/27467
https://github.com/ros/rosdistro/pull/27466

AFAIK there isn't a fixed release schedule for the official ROS repositories, but they usually push out new releases every 3-4 weeks.  It looks like Dashing & Eloquent were both updated at the end of October and Foxy was updated early November, so it should be soon.

Issue #606, opened 2020-10-27
Make swri_serial_util::SerialPort virtual


RyanDelGizzi:
I would like the swri_serial_util::SerialPort to have its functions declared virtual to allow for overrides in a child class for testing purposes.

PR #607, opened 2020-10-27
Make SerialPort's functions virtual

RyanDelGizzi:
In response to issue #606


pjreed:
I think the CI build failures here are the build farm's fault, everything looks good to me locally.

Issue #609, opened 2020-11-09
ROS Foxy: Calling ageSeconds() on subscriber throws exception


mschickler:
When subscriber.ageSeconds() is called, the following exception is thrown:

terminate called after throwing an instance of 'std::runtime_error'
what():  can't subtract times with different time sources [2 != 1]

PR #610, opened 2020-11-09
Fix calling ageSeconds() on subscriber throws exception (#609)

mschickler:
The original code used rclcpp::Clock::now() to retrieve the time, which returns a clock of type RCL_SYSTEM_TIME.  All other code is using rclcpp::Node::now() which returns a clock of type RCL_ROS_TIME.  The two time sources are not compatible and cannot be compared or operated on together.  This change pushes the code down into the impl where the node handle is available for calling now() when now is not provided as an argument.

It is worth noting that calling age(node->now()) instead of age() is not even possible as a workaround since that will result in another exception due to the passed time object source not being compatible with rclcpp::Time(0,0).  Similar "what" text as the original problem, just having to do with comparing to unlike time sources.


mschickler:
Closing.  Will re-open.

PR #613, opened 2020-11-10
Fix compilation warnings in ROS Rolling

pjreed:
Signed-off-by: P. J. Reed <preed@swri.org>


danthony06:
I think the newest merge I did to pull in support for Galactic has fixed these. I think I'm going to create a rolling branch anyway, so that we can track it there.

PR #615, opened 2020-11-15
Changed swri_yaml_util::Clone to return std::unique_ptr

rjb0026:


pjreed:
Switching away from a deprecated type is good, although why set swri_transform_util to compile with C++11?  ROS Melodic and Noetic both target C++14 by default, so I'm not sure we want to intentionally target an older standard unless it's necessary.

rjb0026:
I made that change just to get the kinetic build to compile. I can update the flags to c++14, but I remember reading somewhere that c++11 is the latest recommended standard version that should be used with kinetic because its not a guarantee that every platform that kinetic is built for comes with a compiler that supports c++14. I don't think it's a bad assumption that most people using kinetic _today_ probably have a compiler that can support c++14.

It would be nice to be able to specify the minimum required version of the standard:
```
set_property(TARGET ${PROJECT_NAME} PROPERTY CXX_STANDARD 14)
target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_11)
```
but that feature isn't available until cmake 3.8

pjreed:
Hmm.  I'd prefer to only force the C++ version in Kinetic and just let it use the default in later distros, but since Kinetic only has cmake 3.5 by default, I don't see a *great* way to do that.  It's a little kludgey since it depends on an environment variable, but you could probably do something like:
```cmake
if("$ENV{ROS_DISTRO}" EQUAL "kinetic")
  set(CMAKE_CXX_STANDARD 11)
  set(CMAKE_CXX_STANDARD_REQUIRED ON)
endif()
```

matt-attack:
This appears to break projects which depend on this library but are not set to build using C++11. I recommend using an ifdef on ROS_VERSION_* to switch for melodic or greater.

rjb0026:
@matt-attack I've been running this with kinetic and melodic for a while now after following your suggestion for an ifdef on the ROS_VERSION. Would you give it another review?

swri_yaml_util/CMakeLists.txt
```diff
@@ -32,10 +35,14 @@ include_directories(SYSTEM
   ${YamlCpp_INCLUDE_DIR}
 )
 
+if ($ENV{ROS_DISTRO} STREQUAL "kinetic")
```
matt-attack, line 38: I dont think we should be setting the global CMAKE_CXX standard here, or even changing it for kinetic at all in this case.

swri_transform_util/CMakeLists.txt
```diff
@@ -60,6 +60,11 @@ include_directories(SYSTEM
   ${PROJ_INCLUDE_DIRS}
 )
 
+if ($ENV{ROS_DISTRO} STREQUAL "kinetic")
+  set(CMAKE_CXX_STANDARD 11)
```
matt-attack, line 64: This should be set per target, not globally

rjb0026: CMAKE_CXX_STANDARD is global per project, so setting the standard for every target in the package does effectively the same thing I believe. I was just trying to save a few keystrokes.

I'm happy to make the change just to be consistent stylistically across all of our other packages.

swri_yaml_util/CMakeLists.txt
rjb0026: You're right that now with the ifdef on the pointer type, I don't think there are any >=c++11 features for this library. I was originally trying to be consistent with the standard setting across the whole marti_common "stack" so that a kinetic user that wants to link to multiple marti_common libraries does not have to worry about ABI issues.

I will go ahead and set it per target, because I still think the whole stack should be >=c++11 compatible.

matt-attack: Well, the issue is we aren't really just making it C++11 compatible, we're potentially breaking backwards compatibility as well. It's probably fine, but it would be a change from previous.

rjb0026: I dove down the rabbit hole a bit to try and find out why we weren't alread having ABI problems with existing c++11 applications that link with swri_yaml_util (assuming they had binaries or built it from source using the c++03 standard flags). It seems that the answer is [dual ABI](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html). xenial / kinetic is a somewhat special release in that *most* users are getting binaries built with gcc >= 5.2 and the macro `_GLIBCXX_USE_CXX11_ABI=1` as the default setting. The use of `std::string` throughout the swri_yaml_util and yaml-cpp library APIs worried me a bit since that was one of the major ABI breaking changes from c++03 to c++11, but I think this is exactly where these default settings are saving us.

I've experimented a bit with linking swri_yaml_util with an executable that uses a couple of the APIs that require `std::string` and set every permutation of the standard properties, turned on `-Wabi`, and manually set `-D_GLIBCXX_USE_CXX11_ABI=0`. Manually setting the macro was the only way I could get the linker to fail. All of that to say: for most kinetic users, it probably doesn't matter.

Kind of a moot point I think because kinetic is no longer receiving updates and the users that would apply this patch in kinetic would likely build from source, it wouldn't get rid of the `std::auto_ptr` warning for them, and hopefully they'll be moving on to a more modern distribution soon:)

I will remove the standard spec properties from this package to reduce the chance of something breaking if someone is going to build with a super old or different compiler toolchain.

rjb0026:
> I know this is pretty old, but since Kinetic is now EOL, could we go back to the simple change, and just automatically target C++14?

I think we should still explicitly have a case for kinetic to set c++11 and then let newer distros default the c++11 standard version. We are still supporting some systems in kinetic and I wouldn't want those (which build from source) to break, but it would be fantastic to get rid of the pervasive cmake warnings.

danthony06 (approved):

PR #617, opened 2021-03-30
Fix usage of yaml_include_dirs

wolfv:
This fixes the build in the RoboStack project: https://github.com/robostack/ros-noetic


matt-attack:
Looks good to me, and doesn't break anything on my end.

PR #618, opened 2021-04-15
[ROS 2] Update swri_transform_util CMakeLists.txt.

sageig:
ROS 2 add_library doesn't defaults to STATIC flag. Updated to specify SHARED.


matt-attack:
This looks reasonable. @danthony06 what do you think?

danthony06 (approved):

PR #623, opened 2021-07-19
Modify swri_math_util to add shared library instead of static library

vknisley-swri:
From what I have seen in a project that I am working on, making the library shared should facilitate linking with other shared libraries.

The whitespace change in line 12 is because I have VSCode set up to remove trailing whitespaces when saving a file.

### DIFF
--- a/swri_math_util/CMakeLists.txt
+++ b/swri_math_util/CMakeLists.txt
@@ -9,9 +9,9 @@ find_package(rclcpp REQUIRED)
 
 # The Boost Random library headers and namespaces changed between version
 # 1.46 and 1.47
-find_package(Boost COMPONENTS random) 
+find_package(Boost COMPONENTS random)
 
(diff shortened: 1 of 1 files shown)


danthony06 (approved):

matt-attack:
Looks good to me. I'm still surprised the default is static libraries in ROS 2.

danthony06:
@matt-attack any idea why these builds are failing? I can't merge until they run.

matt-attack:
No idea. It seems like two of the CI builds didnt run at all or somehow died without reporting.

danthony06:
Hmm, looks like maybe the Galactic and Rolling releases built, but are marked as failed. The Dashing and Eloquent releases are still waiting to run. I suspect this may be because we haven't released for Galactic and Rolling, and the D&E releases are EOL.

PR #626, opened 2021-09-16
Fix template spacing that makes some compilers sad

matt-attack:


matt-attack:
@danthony06 apparently the last thing made some kinetic builds sad. This should fix it.

danthony06 (approved):

matt-attack:
Thanks!

PR #627, opened 2021-09-16
Remove C++11 usages and allow using "~" as a namespace instead of just "~/"

matt-attack:
Realized a few non-C++11 packages needed to use this. This removes any C++11 features.


danthony06:
Could @daniel-dsouza take a look too?

PR #631, opened 2021-10-11
Add support for working with Path and Plan messages

matt-attack:
@daniel-dsouza


matt-attack:
> do we need to wait until marti_messages gets merged before merging this?

Yeah, we should

PR #637, opened 2021-11-17
Fixing warning about deprecated header

danthony06:
Fixes warning about deprecated header.


danthony06:
Also address this, starting in the Foxy release: https://github.com/swri-robotics/marti_common/pull/613

PR #640, opened 2021-11-18
Set param get param support

rookie80:


swri_roscpp/include/swri_roscpp/node_handle.h
```diff
@@ -202,6 +202,34 @@ class NodeHandle
...
   }
 
+  inline bool getParam(const std::string& name, std::vector<std::string>& value,
+    const std::string description)
+  {
+    std::string real_name = resolveName(name);
+    bool set = nh_->pnh_.getParam(real_name, value);
+    ROS_INFO("Read parameter %s", name.c_str());
+
+    if (shouldAddParameter(real_name))
```
danthony06, line 212: If the parameter isn't set, should we even run all of this code?

swri_roscpp/include/swri_roscpp/node_handle.h
```diff
@@ -890,6 +918,14 @@ bool getParam(swri::NodeHandle& nh,
   return res;
 }
 
+template<typename T>
+void setParam(swri::NodeHandle& nh,
+  const std::string name,
```
danthony06, line 923: Should we pass ```name``` by reference?

rookie80: I think so, the NodeHandle setParam function passes name by reference.

swri_roscpp/include/swri_roscpp/node_handle.h
rookie80: That is a valid point, the XmlRpcValue getParam function has the same structure as this vector<std::string> getParam, so I am unsure what we want to do here.

rookie80: Nevermind ^, we do want to run all of this code. This swri implementation of node_handle documents the parameters we are using in our code. We would especially want to document params if they aren't set, so that when the rosman check script (the issue you merged earlier) runs, it can see that there is a parameter that is documented but not set, and output that.

matt-attack:
@danthony06 this should be good to go since the necessary changes were made

swri_roscpp/include/swri_roscpp/node_handle.h
danthony06: Can you update it to pass by reference then?

matt-attack: It looks like it has been updated. This comment is on an old version, the current one seems to have it

PR #641, opened 2021-11-19
Add advertise later option for image transport topics.

sageig:


matt-attack:
@danthony06 how does this look? It looks good to me

PR #643, opened 2022-01-04
added xlmrpc support for rosman check, changed output for rosman check

rookie80:


rosman/src/rosman/__init__.py
```diff
@@ -509,8 +509,9 @@ def compare_param(rosmaster, node_name, yaml=False, output_file=sys.stdout, reve
     doc_server_array = []
     param_server_array = []
     unused_param = []
-
+    #print("compare_param top")
```
danthony06, line 512: Please remove dead code

danthony06 (changes requested):

rosman/src/rosman/__init__.py
rookie80: It is done

Issue #644, opened 2022-01-10
local_xy_util doesn't choose shortest distance when crossing longitude of 180 degrees


tothgl:
If the local xy origin and the current location are on opposite sides of the longitude 180/-180 line, the function LocalXyWgs84Util::ToLocalXY() calculates the position difference the long way around the globe instead of the short distance across the 180 degree boundary.

pjreed:
Good catch, this is probably something @kriskozak will want to know about.

tothgl:
FYI, I forked the repo and checked in the code changes and test cases I made on the "PR644" branch. I can create a pull request if desired.

danthony06:
We're always happy for pull requests.

tothgl:
I made my changes to version 2.14.2 which was the version released with kinetic and noetic. 
It looks like master and 2.15.0 have a lot of changes (none to the two files I changed) and I haven't been able to get either of them to compile yet. Are they targeted for ROS2?
Should my pull request be against 2.14.2 or master (or something else)?

danthony06:
Right now, we're releasing ROS1 based on the master branch, and ROS2 based on the ros2-devel branch. Sorry for the confusion about the releases, but so far we've been trying to release both versions out of one repo.

danthony06:
If you can apply the fix to whichever branch you feel most comfortable with, we can port it over to the other branch.

kriskozak:
> Good catch, this is probably something @kriskozak will want to know about.

Thanks for the heads up. The issue at 180 degrees is not a surprise, and I don't recall exactly, but it may have been known of previously and ignored because there aren't many practical applications that would trigger the issue. Given that there is a simple fix, though, it makes sense to just fix it.

Out of curiosity @tothgl, what led you to discover this issue, and does it involve a boat :-)?

While we're at it, I'll note that there will also be issues with this Cartesian mapping near the poles, but I don't think there is a similarly simple fix for that.

tothgl:
@kriskosak No, I'm afraid there was no boat involved nor did I get to go to Fiji :-) I was just writing test cases for a gps driver using the worst cases I could think of. 
I thought about the pole issue and didn't have any ideas for that either. 
I wanted to let the issue go, but it kept nagging at me...

danthony06:
It is now fixed in ROS 2 as well.

PR #645, opened 2022-01-12
Allowing wrapping across longitude 180/-180 line. (Issue #644)

tothgl:
I think this fixes the wrapping issue. 
Let me know if there is anything further I should do.
Thanks.
Gabe

### DIFF
--- a/swri_transform_util/src/local_xy_util.cpp
+++ b/swri_transform_util/src/local_xy_util.cpp
@@ -272,7 +272,11 @@ namespace swri_transform_util
       double rlat = latitude * swri_math_util::_deg_2_rad;
       double rlon = longitude * swri_math_util::_deg_2_rad;
       double dLat = (rlat - reference_latitude_) * rho_lat_;
-      double dLon = (rlon - reference_longitude_) * rho_lon_;
+      double dLon = (rlon - reference_longitude_);
+      // Check for case where the shortest distance is across the date line
+      if (dLon > swri_math_util::_pi) dLon -= swri_math_util::_2pi;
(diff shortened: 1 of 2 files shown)


tothgl:
Some of the automated tests failed.
I saw something similar when I tried to build the master branch under noetic. 
I ended up having to manually clone the swri-robotics/marti_messages repo. (I previously tried running rosdep, but it gave me an error with marti_introspection_msgs.)
I will investigate...

swri_transform_util/src/local_xy_util.cpp
```diff
@@ -272,7 +272,11 @@ namespace swri_transform_util
       double rlat = latitude * swri_math_util::_deg_2_rad;
       double rlon = longitude * swri_math_util::_deg_2_rad;
       double dLat = (rlat - reference_latitude_) * rho_lat_;
-      double dLon = (rlon - reference_longitude_) * rho_lon_;
+      double dLon = (rlon - reference_longitude_);
+      // Check for case where the shortest distance is across the date line
```
kriskozak, line 276: Just a minor point: I think you mean the 180th meridian or the antimeridian (I had to look that up to be sure) rather than the date line, which only partially coincides with the antimeridian: https://en.wikipedia.org/wiki/International_Date_Line.

kriskozak:
This fix looks good to me.

danthony06 (approved):
I made a couple of small style changes, but looks good to me as well.

swri_transform_util/src/local_xy_util.cpp
tothgl: Yes. My use of the term dateline was not accurate.

PR #649, opened 2022-02-10
swri_transform_util: try to find PROJ with CMake

lopsided98:
PROJ provides CMake config files when installed with CMake (rather than autotools, as is done in Debian/Ubuntu). Versions older than 8 (which isn't compatible with `swri_transform_util`) don't install a pkg-config file when installed with CMake. Therefore, the build currently fails to find PROJ on systems where it is install with CMake to a directory other than `/usr`.

This patch causes all three methods to be tried in sequence, so should be compatible with all these possibilities. I don't see why using `find_library()` would ever be necessary, but I kept it to maintain compatibility.


danthony06:
Great, thanks!

PR #654, opened 2022-04-18
Add missing header.

### LINKED ISSUE #653: ros/ros.h must be included before swri_roscpp/topic_service_client.h
Hi,

I wrote a really simple node that only calls a topic_service_client. On the initial implementation, I listed the headers in the following order:
```                                                                                                                                                                                                                                                                                                                                                                              
#include <swri_roscpp/topic_service_client.h>                                                                                                                                                                                                                                                                                                                                               
#include <ros/ros.h>        
```
which failed to compile, with the error:
```
error: 'ros::this_node' has not been declared                                                                                                   
     name_ = client_name.length() ? client_name : (ros::this_node::getName() + random_str);                                                                                                                                                                                              
```

I believe that `ros/this_node.h` needs to be included in `#include <swri_roscpp/topic_service_client.h>`.

jpace121:
Adds the `ros/this_node.h` header to `topic_service_client.h`, which should fix the issue pointed out in #653.

Fixes #653.

### DIFF
--- a/swri_roscpp/include/swri_roscpp/topic_service_client.h
+++ b/swri_roscpp/include/swri_roscpp/topic_service_client.h
@@ -30,6 +30,7 @@
 #define SWRI_ROSCPP_TOPIC_SERVICE_CLIENT_H_
 
 #include <ros/node_handle.h>
+#include <ros/this_node.h>
 
 #include <boost/thread/mutex.hpp>
 #include <boost/uuid/random_generator.hpp>


jpace121:
I don't think this change is the cause of the build failures in the tests. I'm seeing the same issue on master locally. I'm working from a different branch locally that I've applied this change to, which is how I was able to test things.

danthony06:
Yeah, the build is failing because it's looking for ```marti_introspection_msgs```. I'll doublecheck if this was released for the builds that are failing.

PR #659, opened 2022-04-28
Initialize members with 'shared_from_this()' after constructor

agyoungs:


matt-attack:
To a bit pedantic, I'll note all of these should actually be unique_ptrs because they only have one owner.

agyoungs:
This could potentially be handled with states such as the life cycle in managed nodes handles. This would be a larger PR if we converted everything to managed nodes

PR #663, opened 2022-05-13
Cleaning up build warnings related to durations, parameter types, and…

danthony06:
… string formatting


danthony06:
@agyoungs would you mind taking a quick look at this? Nothing too risky, just fixing a bunch of build  warnings that were making it tough to catch real problems.

agyoungs (changes requested):
All looks good. I'd just change to the static call of zero

Issue #647, opened 2022-01-28
swri_roscpp build issue in ROS Noetic (Python3 compatibility issue)


vknisley-swri:
This was discovered when trying to build some packages in ROS Noetic.

The `service_splitter.py` script opens files in binary mode for reading and writing. Changes between Python2 and Python3 have made it so that only bytes, not strings, can be read from/written to files in such cases, and the conversion is not automatic. [This is a Stack Overflow question concerning a similar situation.](https://stackoverflow.com/questions/33054527/typeerror-a-bytes-like-object-is-required-not-str-when-writing-to-a-file-in) The issue can be fixed by either converting the strings to bytes or by not opening the files in binary mode.

TyHowellWork:
@vknisley-swri this should now be fixed by [this commit](https://github.com/swri-robotics/marti_common/commit/d79e6bc600027b635f6552f55f5488a5494ae06d).

PR #668, opened 2022-05-27
open service files in text mode

TyHowellWork:
As mentioned in #647, swri_roscpp service splitter was opening service files in binary mode but parsing as strings, which Python3 does not [allow](https://docs.python.org/3/howto/pyporting.html):

> Under Python 3, binary files and text files are clearly distinct and mutually incompatible; see the [io](https://docs.python.org/3/library/io.html#module-io) module for details. Therefore, you must make a decision of whether a file will be used for binary access (allowing binary data to be read and/or written) or textual access (allowing text data to be read and/or written).

This fixes that, allowing services to be split in ROS Noetic and will also work with previous releases targeted by the master branch. The ros2 branches appear to have this fix already.


danthony06 (approved):

danthony06:
This looks good, but let me figure out how to branch this right. I think we're going to need a branch for Melodic until it goes EOL, since it's using Python 2.7.

TyHowellWork:
> 

I tested the changes with Melodic as well as Noetic, it should be backwards compatible as long as the service_splitter is not opening binary files (not sure why it would unless I am overlooking it).

PR #670, opened 2022-05-31
Added null check to swri::TransformManager::Initialize

rjb0026:


danthony06:
We shouldn't be initializing a listener inside the class. See the comments in 616 about why this is bad.

rjb0026:
Ok, well maybe we can just add parity for the logging and fail out added to `Initialize` in the ROS 2 version? Like was added [here](https://github.com/swri-robotics/marti_common/pull/666/files#diff-dd604c9518a68b4896ea73a6173896f2f1f8910d478fcb4af7e53558cae48a78R84). I think it would probably be more useful to a user to throw an exception in this case as the "fail out".

If we don't return or throw an exception for this case (a nullptr value being passed explicitly to `Initialize`), then there is nothing stopping one of the transformer implementations from possibly initializing the transform listener or using the uninitialized listener, causing a crash in a weird spot. I found this problem from a crash in the wgs84_transformer [where it assumes the transform listener passed in to its `Initialize` method is valid](https://github.com/swri-robotics/marti_common/blob/master/swri_transform_util/src/wgs84_transformer.cpp#L113). I suppose I can fix that while I'm at it.

swri_transform_util/src/wgs84_transformer.cpp
```diff
@@ -110,7 +110,12 @@ namespace swri_transform_util
     if (local_xy_util_->Initialized())
     {
       std::string local_xy_frame = local_xy_util_->Frame();
-      if (tf_listener_->frameExists(local_xy_frame))
+      if (!tf_listener_)
+      {
+        ROS_WARN("Cannot initialize a Wgs84Transformer with an invalid transform listener.");
+        initialized_ = false;
```
danthony06, line 116: I'm not sure if changing the initialized_ value here is a good,

danthony06 (changes requested):

danthony06:
If we're going to make this change, i think it would be nice to bring in the change from the ```ros2-devel``` branch that lets us pass in the tf_listener to the constructor and conditionally call the Initialize() function based on that.

swri_transform_util/src/wgs84_transformer.cpp
rjb0026: Yep, I see now that its just going to get overwritten by the return value anyway. I do think the state should `initialized_ == false` if the tf listener is not valid, but I will change that to a local variable that gets returned.

rjb0026:
I think we could definitely add an optional TF listener argument to the constructor that would conditionally call the Initialize method. That would remain backwards compatible. If we copied the ROS 2 API directly and got rid of the default initialization of a TF listener when calling Initialize with no arguments, that would break older code.

danthony06 (approved):

PR #672, opened 2022-06-23
Make Topic Service Client Optionally Spin

danthony06:
Adds the ability to have the topic service client either spin itself, or have an external source of spinning the callbacks.

### DIFF
--- a/swri_roscpp/include/swri_roscpp/topic_service_client.h
+++ b/swri_roscpp/include/swri_roscpp/topic_service_client.h
@@ -87,24 +87,45 @@ public:
 
   bool wait_for_service_nanoseconds(std::chrono::nanoseconds timeout)
   {
+    // Inspired by process in ClientBase::wait_for_service_nanoseconds in client.cpp
     auto start = std::chrono::steady_clock::now();
-    bool conn_available = false;
-    while (
-      !conn_available && 
-      (std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::steady_clock::now() - start) < timeout) &&
-      rclcpp::ok() &&
-      (request_pub_->get_subscription_count() == 0 || response_sub_->get_publisher_count() == 0))
+
+    if (timeout == std::chrono::nanoseconds(0))
     {
-      rclcpp::sleep_for(std::chrono::milliseconds(2));
-      conn_available = true;
+      return false;
     }
 
-    RCLCPP_ERROR_EXPRESSION(
+    std::chrono::nanoseconds time_to_wait = std::chrono::nanoseconds::max();
+    if (timeout > std::chrono::nanoseconds(0))
+    {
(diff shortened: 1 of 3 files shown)


jonselling:
Is there CI on this repo that is run before PRs get merged? Currently building fails for me due to

```
marti_common/swri_roscpp/include/swri_roscpp/topic_service_client.h:246:72: error: missing template arguments before ‘(’ token
  246 |     std::chrono::duration<RepT, RatioT> timeout = std::chrono::duration(std::chrono::nanoseconds(1s)))
```

Building on 20.04 with ROS2 Foxy

danthony06:
The CI has been really flaky lately. I'll try to figure out why you're not
building.

swri_roscpp/include/swri_roscpp/topic_service_client.h
```diff
@@ -29,88 +29,159 @@
...
   std::string name_;
   std::string service_name_;
+  bool internal_spin_;
 
   int sequence_;
 
 public:
-  TopicServiceClientRaw() :
-    timeout_(std::chrono::duration<float>(4.0)),
+  TopicServiceClientRaw(bool internal_spin=true) :
```
jonselling, line 59: I think with addition of this in the constructor then requires a non-default constructor for the `TopicServiceClient` class to control this flag

swri_roscpp/include/swri_roscpp/topic_service_client.h
```diff
@@ -159,16 +233,26 @@ class TopicServiceClientRaw
...
 
 template<class MReq>
-class TopicServiceClient: public TopicServiceClientRaw<typename MReq:: Request, typename MReq:: Response>
+class TopicServiceClient : public TopicServiceClientRaw<typename MReq:: Request, typename MReq:: Response>
 {
 public:
+  template<typename RepT = int64_t, typename RatioT = std::milli>
+  bool
+  wait_for_service(
+    std::chrono::duration<RepT, RatioT> timeout = std::chrono::duration(std::chrono::nanoseconds(1s)))
```
jonselling, line 246: This is the line causing a build error when using the default value. It should be `std::chrono::duration<RepT, RatioT>(...)`

Also, a duration of 1ns intentional? Do we not want to default to 1s or something? If you want the user to pass something in then remove the default argument.

danthony06: Okay, it looks like this is only failing on Foxy, but not Galactic. The duration isn't 1 ns, it's 1 second with a nanosecond precision.

danthony06:
I think all of your issues are addressed in these. It's passing the test cases and the ROS-I CI process, so if there are specific failures with a different test case, knowing that test case would be helpful.

https://github.com/swri-robotics/marti_common/pull/675
https://github.com/swri-robotics/marti_common/pull/674

Issue #679, opened 2022-08-10
wgs84 transformer will not initialize


mdgrogan:
A recent update prevents the wgs84 transformer from setting the initialized flag.

PR #680, opened 2022-08-10
Set initialized_ flag

mdgrogan:
Without this, initialized_ will never be set due to a recent change.


mdgrogan:
This addresses issue #679.

danthony06 (approved):

PR #681, opened 2022-08-29
Rename from rosman tools to swri_cli_tools

danthony06:
@rjb0026 can you take a look at this?

PR #682, opened 2022-09-01
Rename rosman package.

rjb0026:
Renamed rosman package to swri_cli_tools. Leaving rosman tool and python package intact.


rjb0026:
@danthony06 I think this is the way we want to go. As opposed to #681

danthony06:
Yeah, looks good.

PR #685, opened 2022-09-17
Add support for Humble

shrijitsingh99:
Fixes dependencies and build issues while building for Humble
Updates geos library with and uses their new API
Fixes CMake issues

### DIFF
--- a/swri_geometry_util/CMakeLists.txt
+++ b/swri_geometry_util/CMakeLists.txt
@@ -32,7 +32,7 @@ target_include_directories(${PROJECT_NAME}
   ${GEOS_INCLUDE_DIRS}
 )
 target_link_libraries(${PROJECT_NAME}
-  ${GEOS_LIBRARIES}
+  geos
   Eigen3::Eigen
   opencv_core
 )
@@ -68,6 +68,6 @@ ament_export_dependencies(cv_bridge)
 ament_export_dependencies(tf2)
 ament_export_include_directories(include)
 ament_export_libraries(${PROJECT_NAME}
-  ${GEOS_LIBRARIES})
+  geos)
 
 ament_package()
--- a/swri_roscpp/CMakeLists.txt
+++ b/swri_roscpp/CMakeLists.txt
@@ -98,33 +98,33 @@ ament_target_dependencies(timer_test
   std_srvs
   )
 
-# if (BUILD_TESTING)
-#   find_package(ament_cmake_gtest REQUIRED)
-#   find_package(Boost REQUIRED COMPONENTS system)
+if (BUILD_TESTING)
+  find_package(ament_cmake_gtest REQUIRED)
+  find_package(Boost REQUIRED COMPONENTS system)
 
-#   rosidl_generate_interfaces(${PROJECT_NAME}
-#     msg/TestTopicServiceRequest.msg
-#     msg/TestTopicServiceResponse.msg
-#     DEPENDENCIES std_msgs marti_common_msgs
-#   )
+  rosidl_generate_interfaces(${PROJECT_NAME}
+    msg/TestTopicServiceRequest.msg
+    msg/TestTopicServiceResponse.msg
+    DEPENDENCIES std_msgs marti_common_msgs
+  )
 
-#   ament_add_gtest(topic_service_test_server test/topic_service_test.cpp)
-#   rosidl_get_typesupport_target(cpp_typesupport_target "${PROJECT_NAME}" "rosidl_typesupport_cpp")
-#   target_link_libraries(topic_service_test_server "${cpp_typesupport_target}")
-#   target_link_libraries(topic_service_test_server
-#     Boost::system
-#     swri_roscpp__rosidl_typesupport_cpp
-#   )
-#   ament_target_dependencies(topic_service_test_server
-#     rclcpp
-#     marti_common_msgs
-#   )
-#   target_include_directories(topic_service_test_server
-#     PUBLIC
-#     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-#     $<INSTALL_INTERFACE:include>
-#   )
-# endif()
+  ament_add_gtest(topic_service_test_server test/topic_service_test.cpp)
+  rosidl_get_typesupport_target(cpp_typesupport_target "${PROJECT_NAME}" "rosidl_typesupport_cpp")
+  target_link_libraries(topic_service_test_server "${cpp_typesupport_target}")
+  target_link_libraries(topic_service_test_server
+    Boost::system
+    swri_roscpp__rosidl_typesupport_cpp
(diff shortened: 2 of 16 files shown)


danthony06:
Looks really good at first glance, thank you. Let me see if this will work
with Foxy and Galactic, or if I need to permanently branch them.

swri_roscpp/CMakeLists.txt
```diff
@@ -16,7 +16,28 @@ find_package(std_srvs REQUIRED)
 
 # TODO pjr What to do about swri_roscpp-extras.cmake?
 
-### Build Test Node ###
+add_library(${PROJECT_NAME} INTERFACE)
```
pjreed, line 19: This will cause a cmake error if tests are not disabled; a target named `${PROJECT_NAME}` is already defined below at line 105 with `rosidl_generate_interfaces(${PROJECT_NAME}`.  If this needs to be built and installed as a library, I would recommend renaming the test target.

swri_roscpp/CMakeLists.txt
```diff
@@ -88,7 +109,8 @@ if (BUILD_TESTING)
   )
 
   ament_add_gtest(topic_service_test_server test/topic_service_test.cpp)
-  rosidl_target_interfaces(topic_service_test_server ${PROJECT_NAME} "rosidl_typesupport_cpp")
+  rosidl_get_typesupport_target(cpp_typesupport_target "${PROJECT_NAME}" "rosidl_typesupport_cpp")
```
pjreed, line 112: This will fail to compile on ROS Galactic or earlier because `rosidl_get_typesupport_target` does not exist.

You could replace this with something like:
```cmake
if ("$ENV{ROS_DISTRO}" STRGREATER "galactic")
  rosidl_get_typesupport_target(cpp_typesupport_target "${PROJECT_NAME}_test" "rosidl_typesupport_cpp")
  target_link_libraries(topic_service_test_server "${cpp_typesupport_target}")
else()
  rosidl_target_interfaces(topic_service_test_server "${PROJECT_NAME}_test" "rosidl_typesupport_cpp")
endif()
```

Note that `rosidl_generate_interfaces(${PROJECT_NAME}` above would also need to be changed to `rosidl_generate_interfaces(${PROJECT_NAME}_test` to reflect the new target name.

swri_transform_util/src/nodes/dynamic_transform_publisher.cpp
```diff
@@ -36,7 +36,7 @@
 #include <rcl_interfaces/msg/floating_point_range.hpp>
 #include <rcl_interfaces/msg/parameter_descriptor.hpp>
 #include <tf2_ros/transform_broadcaster.h>
-#include <tf2_geometry_msgs/tf2_geometry_msgs.h>
+#include <tf2_geometry_msgs/tf2_geometry_msgs.hpp>
```
pjreed, line 39: This, and all other files that include `tf2_geometry_msgs.hpp`, will fail to compile on ROS Galactic and earlier because there the file is named `tf2_geometry_msgs.h`.  Unfortunately, while that file exists in Humble, including it will print a warning that cannot be suppressed, which is incredibly annoying.

Probably the best way to make this work across different ROS2 releases is to modify the CMakeLists.txt to pull the `ROS_DISTRO` environment variable and set that in a compiler definition, then use preprocessor macros here to conditionally include either `tf2_geometry_msgs.h` or `tf2_geometry_msgs.hpp`.

swri_geometry_util/CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ target_include_directories(${PROJECT_NAME}
   ${GEOS_INCLUDE_DIRS}
 )
 target_link_libraries(${PROJECT_NAME}
-  ${GEOS_LIBRARIES}
+  geos
```
pjreed, line 35: Unfortunately, the libgeos C++ package is broken in Ubuntu 22.04.  This will compile if you have tests disabled, but if you enable tests, or if anything else tries to use this library, it will fail to link like so:
```
/usr/bin/ld: libswri_geometry_util.so: undefined reference to `geos::geom::Coordinate::Coordinate(double, double, double)'
/usr/bin/ld: libswri_geometry_util.so: undefined reference to `geos::geom::Coordinate::Coordinate(double, double, double)'
collect2: error: ld returned 1 exit status
collect2: error: ld returned 1 exit status
gmake[2]: *** [CMakeFiles/test_intersection.dir/build.make:183: test_intersection] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:187: CMakeFiles/test_intersection.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
gmake[2]: *** [CMakeFiles/test_geometry_util.dir/build.make:183: test_geometry_util] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:215: CMakeFiles/test_geometry_util.dir/all] Error 2
gmake: *** [Makefile:146: all] Error 2
```

Looking through the symbols in the library with `nm -D -C /usr/lib/x86_64-linux-gnu/libgeos.so.3.10.2` reveals that there are no constructors for the `Coordinate` class, so it cannot actually be used.

See https://bugs.launchpad.net/ubuntu/+source/geos/+bug/1980147 for some more details.  The upstream GEOS developers have decided that the package is intentionally broken and they're not going to fix it.

Pretty much the only solution for this is to remove the functions that use libgeos or rewrite them to use the C API.

pjreed (changes requested):

swri_geometry_util/CMakeLists.txt
danthony06: Agreed. If we do this, I think we just make it use the C API for all distros.

swri_roscpp/CMakeLists.txt
danthony06: Can also confirm it's broken on prior ROS2 releases.

danthony06:
@shrijitsingh99 did you want to make these changes? Otherwise I can fork your changes and make them.

danthony06:
@pjreed I've started working on the C API switch here: https://github.com/danthony06/marti_common/tree/c_api_switch.

danthony06:
@pjreed My C API branch is now building and passing tests, but I'm not entirely happy with the implementation and how it's handling GEOS contexts.

danthony06:
I've started working on bringing these changes into a branch here: https://github.com/danthony06/marti_common/tree/humble_support

shrijitsingh99:
@danthony06 Thanks for taking over this. I have not have had the cycles to take this further, sorry about that.
This is still a dependency for my project, and have more time now. Let me know if there is anyway I can help out.

danthony06:
@shrijitsingh99 Thank you for all your help on this. I've based a new PR on this over here: https://github.com/swri-robotics/marti_common/pull/691. I'm closing this PR in favor of the other one.

Issue #688, opened 2022-09-29
Switch to GEOS C API in ROS 2


danthony06:
As noted in this GEOS [issue](https://github.com/libgeos/geos/issues/645), the C++ API is broken in the version of GEOS that ships with Ubuntu 22.04, thus breaking this package for ROS Humble and probably later releases, as noted [here](https://github.com/swri-robotics/marti_common/pull/685). The C++ API is considered unstable and can break at any time, but the C API is never supposed to break. Given this  behavior, we should switch to the C API.

PR #689, opened 2022-09-29
Switch to C API

danthony06:
Resolves https://github.com/swri-robotics/marti_common/issues/688. @pjreed or @rjb0026 , would you mind taking a look at this? I added both the normal C API, and the re-entrant C API if we want to get potential speedups in the future.

### DIFF
--- a/swri_geometry_util/src/util.hpp
+++ b/swri_geometry_util/src/util.hpp
@@ -1,6 +1,6 @@
 // *****************************************************************************
 //
-// Copyright (c) 2015, Southwest Research Institute® (SwRI®)
+// Copyright (c) 2022, Southwest Research Institute® (SwRI®)
 // All rights reserved.
 //
 // Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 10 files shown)


swri_geometry_util/src/util.hpp
```diff
@@ -0,0 +1,50 @@
+// *****************************************************************************
```
roger-strain, line 1: Not sure how these are normally structured, but should this `hpp` file be in `src` or `include`?

danthony06: I agree it also feels weird, but that was ROS2's suggested structure for header files that you don't want exported. That file started with helper functions that didn't create a context to work in, so they were only useful for internal usage.

pjreed:
Sorry for not taking a look at this earlier, I've been away from computers for a while -- this looks pretty good to me, although I was just testing it and found a few things that need to be tweaked in order for this to be successfully imported into another package.  I'll submit a PR with a few tweaks...

PR #691, opened 2022-10-07
Add Humble Support

danthony06:
Adds Humble support while maintaining compatibility with Foxy and Galactic.


tiffanyec:
Your edits look fine to me. Also unsure about the CI fail but I'm not familiar enough with the project and tests to really have any insight on that

danthony06:
I'm not sure what's going on with the failed builds. I'm using the same Docker environment for development on different computers, and it works on one but not the other. It seems to consistently fail on Foxy and Galactic, so it's got to be something with the typesupport CMake change.

PR #693, opened 2022-10-10
Fix ament exports

pjreed:
The swri_route_util and swri_transform_util packages needs to export their definitions and include paths properly in order for other components to be able to import them.

Signed-off-by: P. J. Reed <phillipreed@hatchbed.com>


pjreed:
Ok, this should be good now.

PR #694, opened 2022-11-01
Make ROS 2 GetTransform() Timeout Consistent with ROS 1 Timeout

vknisley-swri:
[In ROS 1](https://github.com/swri-robotics/marti_common/blob/dcdab5002de03ea62cdfac350ee09d3df18e4951/swri_transform_util/src/transform_manager.cpp#L310), a timeout of 0.1 seconds is used in one of swri_transform_utils' GetTransform() calls. This was changed to a timeout of 0 in ROS 2. This PR would revert the timeout back to what it was in ROS 1.

This change should be tested more before it is merged in. Also, I noticed that [this timeout was originally 0.1 seconds in ROS 2 as well](https://github.com/swri-robotics/marti_common/commit/eaa8a7440fc01db86cc7d05089496dd10a61418b), so please let me know if this timeout should remain 0.

### DIFF
--- a/swri_transform_util/src/transform_manager.cpp
+++ b/swri_transform_util/src/transform_manager.cpp
@@ -337,7 +337,7 @@ namespace swri_transform_util
     return GetTransform(target_frame,
         source_frame,
         time,
-        std::chrono::milliseconds(0), transform);
+        std::chrono::milliseconds(100), transform);
   }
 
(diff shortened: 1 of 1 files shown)


tiffanyec:
Has been tested and appears functional

danthony06 (approved):

PR #697, opened 2022-11-29
Fixing how the GEOS library is found

danthony06:
When we switched to the GEOS C library in the prior release we did not update the FindGEOS.cmake file, and as a result, it will not find the GEOS library if the C++ version is not installed.


danthony06:
@pjreed I know you looked at the Humble release last time, does this seem okay to you?

Issue #703, opened 2023-02-24
[Humble]: swri_roscpp does not export dependencies correctly


jonselling:
# Description

When building against `swri_roscpp` in ROS2 humble, there is an error that the `diagnostic_updater` is not exported correctly as a dependency (at least for finding the header file). I get this error unless I add the dependency to my own package:

```
install/swri_roscpp/include/swri_roscpp/subscriber.h:34:10: fatal error: diagnostic_updater/diagnostic_status_wrapper.hpp: No such file or directory
   34 | #include <diagnostic_updater/diagnostic_status_wrapper.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

# Environment

Branch: ros2-devel
Distro: humble

# Actual Results

I get a build error against a dependency my package does not use

# Expected Results

That dependency should get passed as an export from swri_roscpp to my package.

# Additional Info

I checked the CMakeLists.txt and it looks like the dependency _should_ get exported, so I started to mess around with the file a little.

I was able to compile with removing any of the calls that use the `${PROJECT_NAME}` interface and then removing the `BUILD_TESTING` block. 

I am unsure if it is because the ${PROJECT_NAME} export data gets written weirdly with the INTERFACE target and then the `rosidl_generate_interfaces` stuff might interfere in the testing part of the CMakeLists.txt.

danthony06:
I'll take a look at what's going on.

danthony06:
@jonselling when you said you removed any of the calls that use the `${PROJECT_NAME}` interface, what did you mean by that?

jonselling:
I commented out these lines in the CMakeLists.txt
```
add_library(${PROJECT_NAME} INTERFACE)

target_include_directories(${PROJECT_NAME} INTERFACE
  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
  "$<INSTALL_INTERFACE:include>"
)

install(TARGETS ${PROJECT_NAME}
  EXPORT export_${PROJECT_NAME}
  ARCHIVE DESTINATION lib
  LIBRARY DESTINATION lib
  RUNTIME DESTINATION lib
  INCLUDES DESTINATION include
)

...

ament_export_targets(export_${PROJECT_NAME} HAS_LIBRARY_TARGET)
```

And then I also commented out everything in the `if (BUILD_TESTING AND ...)` statement

PR #704, opened 2023-03-16
WIP: Fix dependencies

danthony06:
`swri_roscpp` is not exporting its dependencies correctly, as reported in https://github.com/swri-robotics/marti_common/issues/703. This appears to be caused by ament() macros not understanding what an INTERFACE library is. This PR replaces a lot of the ament() calls with direct CMake calls so that dependencies are correctly exported and linked.

### DIFF
--- a/swri_roscpp/CMakeLists.txt
+++ b/swri_roscpp/CMakeLists.txt
@@ -14,16 +14,22 @@ find_package(rosidl_default_generators REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(std_srvs REQUIRED)
 
-# TODO pjr What to do about swri_roscpp-extras.cmake?
 add_library(${PROJECT_NAME} INTERFACE)
+target_compile_features(${PROJECT_NAME} INTERFACE cxx_std_17)
+target_link_libraries(${PROJECT_NAME} INTERFACE
+  ${diagnostic_updater_LIBRARIES}
+  ${rclcpp_LIBRARIES}
+  ${std_msgs_LIBRARIES}
+)
 
 target_include_directories(${PROJECT_NAME} INTERFACE
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
   "$<INSTALL_INTERFACE:include>"
+  ${rclcpp_INCLUDE_DIRS}
 )
 
 install(TARGETS ${PROJECT_NAME}
-  EXPORT "export_${PROJECT_NAME}"
+  EXPORT "${PROJECT_NAME}Targets"
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION lib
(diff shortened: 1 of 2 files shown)


jonselling:
I can give it a look / test tomorrow

danthony06:
> I can give it a look / test tomorrow

Thanks. I just tested it on my system by install `swri_roscpp` to my system-wide ROS install, and then building a fresh copy of `swri_transform_util`, which I think should pull in the header file that was previously causing you problems.

swri_roscpp/CMakeLists.txt
```diff
@@ -14,16 +14,24 @@ find_package(rosidl_default_generators REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(std_srvs REQUIRED)
 
-# TODO pjr What to do about swri_roscpp-extras.cmake?
 add_library(${PROJECT_NAME} INTERFACE)
+target_compile_features(${PROJECT_NAME} INTERFACE cxx_std_17)
```
jonselling, line 18: You set the CMAKE_CXX_STANDARD above for 14. Should that be left as is or changed?

jonselling:
**FWIW, it looks like the dependencies get caught now.**

However, I don't see the benefit of the Interface Library.
I edited the CMakeLists.txt you had to use the correct ament calls, and I removed the whole library interface.

Instead of exporting the library target, I export the include directory. However, this seems to **not** work when trying to pickup the headers downstream. I looked in my install directory, and the files were actually in the correct spot.

Confused, I looked at the INCLUDE_DIRS set for `swri_roscpp` in my package that depends on it, and it looks like two swri_roscpp directories are set:
1. `install/swri_roscpp/include/swri_roscpp`
2. `install/swri_roscpp/include`

However, when I look at the "compile_commands.json" for the files, I see only `-isystem <absolute_path>/install/swri_roscpp/include/swri_roscpp` directory, so when trying to include anything like `<swri_roscpp/swri_roscpp.hpp>` it will not get picked up. I believe this is because of the `rosidl_generate_interfaces` used.

When I remove all of the code in the `if (BUILD_TESTING ...)` statement in `swri_roscpp`, this gets added to the compile command instead: `-I<path>/install/swri_roscpp/include`

danthony06:
I happened to look at the ROS2 sensor_msgs package, and I'm thinking this may be a cleaner approach to what I was trying: https://github.com/ros2/common_interfaces/blob/rolling/sensor_msgs/CMakeLists.txt. It definitely looks like there's some weirdness around interface only libraries and ament, but that approach looks a lot cleaner.

swri_roscpp/CMakeLists.txt
jonselling, line 1: I believe the section

```
install(DIRECTORY include/
  DESTINATION include
)
```

Might actually need to be
```
install(DIRECTORY include/
  DESTINATION include/${PROJECT_NAME}
)
```

If the messages get built.

jonselling:
After doing some digging, I believe that creating the interface library is _probably_ the way to go, mostly because a lot of things in the `ament_cmake` structure seem to be target based.

PR #732, opened 2024-05-30
Increase minimum CMake version to 3.16

daniel-dsouza:
When building with newer (CMake 3.28), the following warning will appear.
```
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

This PR increases minimum CMake version to 3.16 to satisfy the compatibility check in newer CMake versions.

Note that Ubuntu focal (ROS Noetic) packages CMake 3.16.3, so this will be incompatible with the EOL ROS Melodic.

### DIFF
--- a/marti_data_structures/CMakeLists.txt
+++ b/marti_data_structures/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.16)
 project(marti_data_structures)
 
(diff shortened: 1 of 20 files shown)


danthony06:
Thanks. I made a branch called `melodic-eol` before merging this since it is a breaking change.

PR #739, opened 2024-06-25
Replaced local projection math with geographiclib

JayHerpin:
This change replaces the math uses for the local cartesian projection with GeographicLib's implementation. This means implementation is accurate over much longer distances and also brings it in line with implementation of robot_localization (https://github.com/cra-ros-pkg/robot_localization). This means that using a an swri_transform_util local cartesian transformation will return the same results as using the /{to,from}LL service provided by robot_localization.

### DIFF
--- a/swri_transform_util/CMakeLists.txt
+++ b/swri_transform_util/CMakeLists.txt
@@ -43,6 +43,7 @@ if(NOT "${PROJ_FOUND}")
   )
 endif()
 
+# Geographiclib installs FindGeographicLib.cmake to this non-standard location
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "/usr/share/cmake/geographiclib/")
 find_package(GeographicLib REQUIRED)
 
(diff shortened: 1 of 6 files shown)


danthony06:
@JayHerpin Thanks, this looks really interesting and useful. @rjb0026 and I will dig into this.

JayHerpin:
just FYI. I tested the code on iron only. I haven't yet updated to rolling locally. Also, there were a number of unit tests already failing. I didn't resolve any of those, but I did make sure that there were no regressions when running the tests locally.

Additionally, I notice there were no unit tests dealing with a non-zero reference angle, so that functionality is not really tested.

danthony06:
Thanks. I'm guessing the other failing test cases are part of https://github.com/swri-robotics/marti_common/pull/737, which I'm actively trying to resolve.

danthony06:
Sorry for the slow response. We just merged another change that addressed some of the build failures, so now I'm going to merge this in.

JayHerpin:
cheers!

PR #758, opened 2024-12-24
Configure cmake policy CMP0167 for swri_nodelet

### LINKED ISSUE #757: configure cmake policy CMP0167 for swri_nodelet
Configure swri_nodelet to use the FindBoost included with Boost (available since noetic) vs the CMake packaged FindBoost. This will suppress as warning at build triggered every time swri_nodelet is included.

daniel-dsouza:
fixes #757

### DIFF
--- a/swri_nodelet/CMakeLists.txt
+++ b/swri_nodelet/CMakeLists.txt
@@ -1,4 +1,9 @@
 cmake_minimum_required(VERSION 3.16)
+
+if(${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.30.0")
+  cmake_policy(SET CMP0167 NEW)
+endif()
+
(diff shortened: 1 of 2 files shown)


danthony06:
I tried bumping the CMake version to 3.30, but it looks like that's too new for Noetic. It looks like 3.30 is the minimum required for this, according to https://cmake.org/cmake/help/latest/policy/CMP0167.html.

daniel-dsouza:
I've added logic which will only set this policy if the cmake version is greater than or equal to 3.30.

It builds in the `ros:noetic-ros-base` Docker image. Hopefully will pass CI tests now.

daniel-dsouza:
resolves #757

danthony06 (approved):

PR #763, opened 2025-05-09
[kilted] Update deprecated calls to ament_target_dependencies

DLu:
As of the ROS 2 Kilted release [`ament_target_dependencies` is deprecated.](https://docs.ros.org/en/kilted/Releases/Release-Kilted-Kaiju.html#ament-target-dependencies-is-deprecated)

This PR updates the syntax. Caution should be used to ensure that the target branch is not used for other ROS 2 distros.

Note: There's a lot of changes in this PR, and I'm not 100% certain on them all, but it did get things to successfully compile on my machine.

### DIFF
--- a/swri_console_util/CMakeLists.txt
+++ b/swri_console_util/CMakeLists.txt
@@ -8,8 +8,8 @@ find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
 
 add_library(${PROJECT_NAME} SHARED src/progress_bar.cpp)
-ament_target_dependencies(${PROJECT_NAME}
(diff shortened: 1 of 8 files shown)


danthony06:
Thanks. I think we've still got a few build failures due to other Rolling changes that I'm working through here: https://github.com/swri-robotics/marti_common/pull/767. I'm going to merge this and then rebase my work off of this PR.

danthony06:
@DLu it took a bit, but the current `ros2-devel` branch should have the intent of these changes correctly integrated now. Thanks for kicking off the process on this.

PR #781, opened 2025-06-09
Add dependency on libopencv-dev for pkgs directly using OpenCV

### LINKED ISSUE #780: Packages are missing a dependency on OpenCV
`swri_geometry_util` is currently failing to build on the buildfarm:

1. Humble: https://build.ros2.org/view/Hbin_uJ64/job/Hbin_uJ64__swri_geometry_util__ubuntu_jammy_amd64__binary/78/console
2. Jazzy: https://build.ros2.org/view/Jbin_uN64/job/Jbin_uN64__swri_geometry_util__ubuntu_noble_amd64__binary/46/console

In both cases, it seems to fail because OpenCV isn't found, e.g.:

```
23:21:45 CMake Error at CMakeLists.txt:10 (find_package):
23:21:45   By not providing "FindOpenCV.cmake" in CMAKE_MODULE_PATH this project has
23:21:45   asked CMake to find a package configuration file provided by "OpenCV", but
23:21:45   CMake did not find one.
23:21:45 
23:21:45   Could not find a package configuration file provided by "OpenCV" with any
23:21:45   of the following names:
23:21:45 
23:21:45     OpenCVConfig.cmake
23:21:45     opencv-config.cmake
23:21:45 
23:21:45   Add the installation prefix of "OpenCV" to CMAKE_PREFIX_PATH or set
23:21:45   "OpenCV_DIR" to a directory containing one of the above files.  If "OpenCV"
23:21:45   provides a separate development package or SDK, be sure it has been
23:21:45   installed.
```

Looking at the build history (under "Builds" on the left side), it looks like it started failing _after_ version 3.8.1, i.e., 3.8.1 is the last version that built:

1. Humble: https://build.ros2.org/view/Hbin_uJ64/job/Hbin_uJ64__swri_geometry_util__ubuntu_jammy_amd64__binary/
2. Jazzy: https://build.ros2.org/view/Jbin_uN64/job/Jbin_uN64__swri_geometry_util__ubuntu_noble_amd64__binary/
3. (Rolling is fine because [3.8.1 is the current version](https://github.com/ros/rosdistro/blob/8460c96c592458369da490151af7695048db15f8/rolling/distribution.yaml#L3700))

Looking at [changes after version 3.8.1](https://github.com/swri-robotics/marti_common/compare/3.8.1...ros2-devel), this commit seems most suspicious: https://github.com/swri-robotics/marti_common/commit/647fa9b50e4737e4efebc0182ed8476338ae7d2f. `swri_geometry_util` depends on OpenCV since it uses it in its `CMakeLists.txt`, but it doesn't actually declare a dependency on it in its `package.xml` file (`libopencv-dev`). I think it was simply getting it through `cv_bridge`, but that was removed as a dependency in this commit, which explains the build failure now.

The solution is to declare a dependency on `libopencv-dev`. Looking at other packages in this repo, this isn't the only package that directly uses OpenCV but doesn't declare a dependency on it, although in most cases they get it through `cv_bridge`.

christophebedard:
Fixes #780

These packages all use OpenCV directly (`find_package(OpenCV REQUIRED ...)`) but don't declare a dependency on it.

I haven't actually validated that this fixes the buildfarm issue, but I believe it should.

### DIFF
--- a/swri_geometry_util/package.xml
+++ b/swri_geometry_util/package.xml
@@ -20,6 +20,7 @@
   -->
   <depend>eigen</depend>
   <depend>geos</depend>
+  <depend>libopencv-dev</depend>
   <depend>tf2</depend>
 
   <test_depend>ament_cmake_gtest</test_depend>
--- a/swri_image_util/package.xml
+++ b/swri_image_util/package.xml
@@ -20,6 +20,7 @@
   <depend>eigen</depend>
   <depend>image_geometry</depend>
   <depend>image_transport</depend>
+  <depend>libopencv-dev</depend>
   <depend>message_filters</depend>
   <depend>rcl_interfaces</depend>
   <depend>rclcpp</depend>
--- a/swri_opencv_util/package.xml
+++ b/swri_opencv_util/package.xml
@@ -14,6 +14,7 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
   
   <depend>cv_bridge</depend>
+  <depend>libopencv-dev</depend>
   <depend>swri_math_util</depend>
 
   <export>
--- a/swri_transform_util/package.xml
+++ b/swri_transform_util/package.xml
@@ -26,6 +26,7 @@
   <depend>geographiclib</depend>
   <depend>gps_msgs</depend>
   <depend>geos</depend>
+  <depend>libopencv-dev</depend>
   <depend>marti_nav_msgs</depend>
   <depend>proj</depend>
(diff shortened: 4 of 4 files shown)


christophebedard:
The Hbin job for `swri_geometry_util` passes now: https://build.ros2.org/view/Hbin_uJ64/job/Hbin_uJ64__swri_geometry_util__ubuntu_jammy_amd64__binary/80/

PR #784, opened 2025-07-16
switch to using yaml_cpp_vendor

mallanmba:
This change switches swri_transform_util to use yaml_cpp_vendor so that it will use the same version of yaml-cpp as is used by the core ros packages (e.g. [camera_calibration_parsers](https://github.com/ros-perception/image_common/blob/rolling/camera_calibration_parsers/CMakeLists.txt#L19-L20), [rviz](https://github.com/ros2/rviz/blob/rolling/rviz_common/CMakeLists.txt#L56), etc.) This is important for folks who are building from source on platforms that may have a different system version than the version ros wants.
We are on Ubuntu 22.04 which comes with yaml-cpp 0.7 and using ros2 jazzy which wants 0.8. Mixing versions can cause problems downstream.

### DIFF
--- a/swri_transform_util/CMakeLists.txt
+++ b/swri_transform_util/CMakeLists.txt
@@ -20,6 +20,7 @@ find_package(swri_roscpp REQUIRED)
 find_package(tf2 REQUIRED)
 find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
+find_package(yaml_cpp_vendor REQUIRED)
 find_package(yaml-cpp REQUIRED)
 
 # tf2_geometry_msgs 0.12.1 broke API compatibility, so check which version
@@ -76,7 +77,7 @@ target_link_libraries(${PROJECT_NAME} PUBLIC
   opencv_imgproc
   opencv_video
   ${PROJ_LIBRARIES}
-  yaml-cpp
+  yaml-cpp::yaml-cpp
   ${geometry_msgs_TARGETS}
   rclcpp::rclcpp
   swri_math_util::swri_math_util
--- a/swri_transform_util/CMakeLists.txt
+++ b/swri_transform_util/CMakeLists.txt
@@ -21,7 +21,6 @@ find_package(tf2 REQUIRED)
 find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(yaml_cpp_vendor REQUIRED)
-find_package(yaml-cpp REQUIRED)
 
 # tf2_geometry_msgs 0.12.1 broke API compatibility, so check which version
 # we need to use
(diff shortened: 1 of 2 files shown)


swri_transform_util/CMakeLists.txt
```diff
@@ -21,7 +21,6 @@ find_package(tf2 REQUIRED)
 find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(yaml_cpp_vendor REQUIRED)
-find_package(yaml-cpp REQUIRED)
```
mallanmba, line 23: This find_package is needed, see pattern used in other ros packages:
https://github.com/ros2/rviz/blob/rolling/rviz_common/CMakeLists.txt#L56-L57
https://github.com/ros-perception/image_common/blob/rolling/camera_calibration_parsers/CMakeLists.txt#L19-L20
https://github.com/ros2/rosbag2/blob/rolling/rosbag2_storage/CMakeLists.txt#L29-L30

swri_transform_util/CMakeLists.txt
```diff
@@ -77,13 +76,13 @@ target_link_libraries(${PROJECT_NAME} PUBLIC
   opencv_imgproc
   opencv_video
   ${PROJ_LIBRARIES}
-  yaml-cpp::yaml-cpp
```
mallanmba: yaml-cpp::yaml-cpp is the correct target, the modern CMake target sets up include directories and link libraries, see conversation here: https://github.com/ros2/yaml_cpp_vendor/issues/51#issuecomment-2149675576

swri_transform_util/CMakeLists.txt
```diff
@@ -235,6 +233,6 @@ ament_export_dependencies(
   tf2
   tf2_geometry_msgs
   tf2_ros
-  yaml-cpp)
```
mallanmba, line 237: `yaml-cpp` is the correct target to export, my understanding is `yaml_cpp_vendor` sets up the right hints to get the yaml-cpp find_package to find the vendored package, but the actual target is `yaml-cpp`.
In packages.xml, depending on `yaml_cpp_vendor` tells rosdep to look for the vendored package, if `yaml-cpp` is used rosdep will want to use the system package - but that causes problems on platforms that have a different `yaml-cpp` system version than the _vendor package.
Without the export of yaml-cpp, swri_route_util will not be able to find and link against yaml-cpp.
See the exports of rosbag2, they export both yaml-cpp and yaml_cpp_vendor - I don't think it's required to export yaml_cpp_vendor, but it doesn't hurt. Exporting yaml-cpp is required for downstream packages to link properly.

swri_transform_util/CMakeLists.txt
danthony06: It looks like that's only for Jazzy and later, and we'll have to add a switch for Humble.

mallanmba: Ok, I'll put up another PR later today with a switch for the target

PR #785, opened 2025-07-17
add a check for the modern CMake yaml-cpp target

mallanmba:
This adds a check for the modern CMake yaml-cpp target and falls back to YAML_CPP_LIBRARIES if it does not exist, based on comments in [pull 784](https://github.com/swri-robotics/marti_common/pull/784).


danthony06 (approved):

danthony06:
@mallanmba thanks for the PR. I've bloomed these changes to Humble, Jazzy, Kilted, and Rolling as part of v3.8.7.

PR #790, opened 2026-06-19
Lyrical Update

danthony06:
These updates the code to remove calls to deprecated functionality in Lyrical. There's a few common changes:

- Updates to the image_transport::ImageTransport() constructor
- How images are subscribed to and published
- tf2 headers that are renamed to .hpp from .h
- Updated calls to `spin_once` and `spin_some` to use the executor instead
- ament_index_cpp now uses std::filesystem objects instead of strings
- Completely removed swri_roscpp. I don't think anyone is using this, and it's proving increasingly difficult to maintain.


danthony06:
@agyoungs or @DangitBen would you mind taking a look at this? I'm not sure if I'm doing the image_transport stuff the "right" way now. I also couldn't think of doing this in a clean way that would allow us to continue releasing all versions, so we'll probably need a `kilted-devel` branch after this for the pre-Lyrical releases. I think if this is merged in, we'll finally be able to get Mapviz bloomed for Lyrical.

PR #792, opened 2026-07-17
Remove GeographicLib Specific CMake

danthony06:
Forward declaring GeographicLib pointer so we do not need extra cmake to make downstream projects aware of GeographicLib.


jonselling:
FYI I think this PR makes the LocalXyUtil class non-copyable since a unique_ptr is used. Might be worth making your own copy constructor where you move it, or just leave it as is.

I only know this because 3.9.1 was added for Jazzy in the last week or so I believe.

danthony06:
Thanks, I'll make a separate PR to fix this.

danthony06:
@jonselling I added a copy constructor and assignment operator. It's on the `ros2-devel` branch now and bloomed in release 3.11.0.

PR #800, opened 2026-08-29
Adding Transform equality so callers can detect when a transform changes

danthony06:


danthony06:
@DangitBen this helps fix https://github.com/swri-robotics/mapviz/issues/909 cleanly. The change by itself is kind of nice because it's an equality check for a transform, so I think it's worth adding no matter what happens on the Mapviz issue.

